PR #147, opened 2012-08-11
TouchEvent Pooling

joshtynjala:
I have updated TouchProcessor to use pooled touch events. The pooling is exposed as static functions on TouchEvent. I tried to keep it very similar to the starling.events.Event implementation of pooling. The main difference is that doesn't tie into dispatchEventWith() in any way. TouchProcessor directly calls fromPool() and toPool().

I'm seeing a small FPS improvement when scrolling a Foxhole list on my iPad 1. It's only about 1-3 FPS, but I tested the pooled implementation a number of times to be sure, and it is consistently better than the old implementation with TouchEvents that are not pooled.

### DIFF
--- a/starling/src/starling/core/TouchProcessor.as
+++ b/starling/src/starling/core/TouchProcessor.as
@@ -116,9 +116,14 @@ package starling.core
                 // if the target of a hovering touch changed, we dispatch an event to the previous
                 // target to notify it that it's no longer being hovered over.
                 for each (var touchData:Object in sHoveringTouchData)
+                {
                     if (touchData.touch.target != touchData.target)
-                        touchData.target.dispatchEvent(new TouchEvent(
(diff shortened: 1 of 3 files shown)


subdan:
Hello, Daniel. Do you plan to add pooling to TouchEvent in Starling 2.0?

PrimaryFeather:
Ah, now this is an issue I totally forgot about. Thanks for the reminder — I can't promise anything, but it might happen. :wink:

PrimaryFeather:
The latest code was already only using one TouchEvent per frame; thus, pooling was not even necessary: there's only a single `TouchEvent` instance now that's reset and reused over and over.

joshtynjala:
Nice. Thanks, Daniel!

subdan:
Great! Thanks.

PrimaryFeather:
You're welcome! And @joshtynjala: thanks for the original pull request, and sorry that it took ... wow ... 4 years! :wink:

PR #302, opened 2013-03-12
Optimizing VertexData

milk531:
I'v recently read the article "Boosting Starling 1.3 performance by 50%" and try to do the optimization as its said, and I also make VertexData's constructor a little faster. I does some tests and it seems fine.


PrimaryFeather:
(Cleaning up on old issues and pull requests.)

Starling 2.0 contained a complete rewrite of the VertexData class, so this doesn't apply any longer.

PR #454, opened 2013-12-21
FragmentFilter changes and new Bloom Filter

### LINKED ISSUE #391: FragmentFilters Interface
Hi Daniel,

the last weeks I got lost somehow in writing FragmentFilters and I recognized that some Filters will only be possible with overriding the render function. (That recently came up with a 2.5D Filter, ... I know, bad idea ... but will also be an issue with the BloomFilter I had in mind)

Unfortunately overriding is not working well, since a lot of private variables/functions are used in your render/renderPasses functions. My workaround was disposing the filter right in the subclass's constructor and adding almost a complete copy of the FragmentFilter super class to my subclass, which is obviously bad. But I couldn't find a better way to make it work without customizing starling, which would make the filters useless for others.

So I'd like to ask whether you could implement a IFragmentFilter interface to Starling 1.4 or at least change the private functions/vars to protected, which would also be working, but as I guess be less efficient. What ever you like best. That would be really awesome.

Thanks in advance

Michael

mountainstorm:
I've created a BloomFilter (like the fog filter in Pixelmator); basically a gaussian blur on a new layer, blended with the lighten mode.

To allow this I've had to make some changes to the FragmentFilter.  One (changing the protection on calculateBounds) is optional but useful if you ever need to do size calculations in a FragmentFilter subclass.

The other is to allow rendering to a filter, rather than the backbuffer; see here for an alternative solution http://forum.starling-framework.org/topic/mist-or-glow-filter

Overall I think this is a better solution as its more de-coupled.  In addition don't thing the (removed) assertion is valid -  Stage3D is quite happy to render AGAL programs to textures.

I think this also resolves #391 (its a bloom filter :); although I'd echo the comments there on making it easier to extend FragmentFilter as, although it's not a major issue given this implementation, it was a pain in the ass while trying other solutions.

### DIFF
--- a/starling/src/starling/filters/BloomFilter.as
+++ b/starling/src/starling/filters/BloomFilter.as
@@ -1,0 +1,112 @@
+// =================================================================================================
+//
+//	Starling Framework
+//	Copyright 2012 Gamua OG. All Rights Reserved.
+//
+//	This program is free software. You can redistribute and/or modify it
+//	in accordance with the terms of the accompanying license agreement.
+//
+// =================================================================================================
+
+package starling.filters
+{
+	import flash.display3D.Context3D;
+	import flash.display3D.Context3DBlendFactor;
+	import flash.display3D.Context3DProgramType;
+	import flash.display3D.Program3D;
+	import flash.display3D.VertexBuffer3D;
+	import flash.display3D.IndexBuffer3D;
+	import flash.geom.Rectangle;
+	import starling.filters.BlurFilter;
+	import starling.filters.FragmentFilter;
+	import starling.textures.Texture;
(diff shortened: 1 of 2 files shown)


PrimaryFeather:
I know this pull request is already quite old, but I just wanted to update this:
Since around AIR 16, it's officially possible to stack filters. In fact, I'm now relying on this in Starling 2.0, which greatly enhanced the possibilities of filters.

Thus, if you're still using Starling, it should finally be possible to add that filter to Starling without having to modify Starling in any way. You would need to rewrite it (tutorial [here](http://wiki.starling-framework.org/manual/creating_a_custom_filter)), though.

PR #540, opened 2014-05-08
juggler.removeDelayedCall

lagoonSoft:
added `juggler.removeDelayedCall(call)` to be able to remove a previously added `DelayedCall` with `juggler.delayedCall()` method or a manually added `DelayedCall` to the juggler. 
I just cloned the `juggler.removeTweens()` function and added a getter to the actual call in the `DelayedCall`


PrimaryFeather:
I actually added this quite a while ago, but apparently forgot to close the pull request. Sorry for that, and thanks again for your help!

PR #561, opened 2014-06-10
-changed some private vars to protected to be able to extend QuadBatch

nodrock:
I need to extend QuadBatch and want to be able to batch it with other QuadBatches.


PrimaryFeather:
In Starling 2, you have more access to those property (via the new `MeshStyle`), and through the `VertexData` instance. So this shouldn't be needed any longer.

PR #564, opened 2014-06-20
Fixed > AssetManager: Texture removed from mTextures when adding a font...

Nayael:
Fixed > AssetManager: Texture removed from mTextures when adding a font or atlas

(Problem explained on the forum: http://forum.starling-framework.org/topic/assetmanager-texture-atlas-issue-since-15#post-64980)


PrimaryFeather:
I actually added this a while ago, but apparently forgot to close the pull request. It's now working just as you expected it.

PR #577, opened 2014-07-05
Minor performance improvement on handling events. (one if removed) :

MindScriptAct:
- removed check if starling is started.
- ensured events are not listened if starling is not started.


MindScriptAct:
Oh.. nevermind.

PR #592, opened 2014-08-04
Added TravisCI build script

vpmedia:
Continuous builds using Travis-CI service:
https://travis-ci.org/vpmedia/Starling-Framework/builds/31642080

I think this would be useful to validate other feature requests too, using automated scripts.

Please note that the ant build script assembles the SWC but does fail to run the ASDoc script, because OSX brew package manager has only AdobeAIR SDK support (Apache Flex SDK is problematic to install headless) so the Ant build script needs some refactoring.
Not sure that that should be done in a separated feature branch or in this one.

Cheers!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+git: 
+    depth: 1
+
+language: objective-c
+
+before_script:
+    - brew update
+    - brew install ant
+    - brew install adobe-air-sdk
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
I'm sorry to bring this up again, but builds are failing again at the moment. Could you maybe look into it when you've got a little time?

    Error: No available formula for adobe-air-sdk 

Thanks a lot in advance!!

PR #628, opened 2014-10-11
Added setRenderTargets method to the RenderSupport

Varnius:
As the new Standard profile added the possibility to set multiple render targets I think it would be nice to have a way to acces this functionality from RenderSupport. While MRTs aren't really used by Starling in any way, I think it may prove handy to extension developers since everything related to RTs would be accessible through RenderSupport without mixed calls either to Context3D instance or RenderSupport depending on which version of setRenderTarget you need.

I've also added an optional parameter for enabling depth and stencil buffers if rendering to a texture.

### DIFF
--- a/starling/src/starling/core/RenderSupport.as
+++ b/starling/src/starling/core/RenderSupport.as
@@ -226,20 +226,57 @@ package starling.core
             setRenderTarget(target);
         }
 
-        /** Changes the the current render target.
-         *  @param target       Either a texture or 'null' to render into the back buffer.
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
We discussed this somewhere else (maybe in the forum?), and I think the consensus was that its best if developers use the context method directly, instead of adding this to the `RenderState` of Starling 2; mainly because setting an additional render target also means that all fragment shaders would need to change (they must write to that output, then) - which means one has to do a lot of custom context calls, anyway.

Feel free to reopen this if my memory fails me on this! :wink:

PR #642, opened 2014-11-08
Bugfix: RenderTexture's draw and drawBundled methods threw exceptions if context was lost.

stepan-beresnev:
It's basically the same issue as was discussed here [627](https://github.com/Gamua/Starling-Framework/pull/627).


stepan-beresnev:
Hello,

Not sure, if you did it intentionally, but you didn't copy
if (!Starling.current.contextValid) return;
on the top of RenderTexture's render function.

PrimaryFeather:
That's intentional! "render" will never be called alone; it's always called from within "renderBundled". So it's enough to check the context in the latter method.

PR #674, opened 2015-02-08
Open mCall in DelayedCall

nomikovsupe:
Sorry my english.

My class with call getter:

```as3
public class InstantDelayedCall implements IAnimatable
{
	public function InstantDelayedCall(call:Function, delay:Number)
	{
		_delayedCall = new DelayedCall(call, delay);
		_juggler.add(_delayedCall);
	}
	
	public function advanceTime(time:Number):void
	{
		if (_delayedCall.currentTime == 0) _delayedCall.call();
		_juggler.advanceTime(time);
	}
	
	private var _juggler:Juggler = new Juggler();
	private var _delayedCall:DelayedCall;
}
```

without call getter:

```as3
public class InstantDelayedCall implements IAnimatable
{
	public function InstantDelayedCall(call:Function, delay:Number)
	{
		_call = call;
		_delayedCall = new DelayedCall(call, delay);
		_juggler.add(_delayedCall);
	}
	
	public function advanceTime(time:Number):void
	{
		if (_delayedCall.currentTime == 0) _call();
		_juggler.advanceTime(time);
	}
	
	private var _call:Function;
	private var _juggler:Juggler = new Juggler();
	private var _delayedCall:DelayedCall;
}
```

I want to store "call" only in one place.
Maybe "DelayedCall.mArgs" should be opened too.

Sorry if i sound rude. I know english bad.

### DIFF
--- a/starling/src/starling/animation/DelayedCall.as
+++ b/starling/src/starling/animation/DelayedCall.as
@@ -86,6 +86,9 @@ package starling.animation
             return mRepeatCount == 1 && mCurrentTime >= mTotalTime; 
         }
         
+        /** The function, that will be called */
+        public function get call():Function { return mCall; }
+        
         /** The time for which calls will be delayed (in seconds). */
         public function get totalTime():Number { return mTotalTime; }
         


PrimaryFeather:
Hi nomikovsupe, thanks for the pull request!
I'll consider making these properties available; however, if the reason for your creating the "InstanstDelayedCall" class is to make it easier to delay a function call, you could also use the shortcut on the "Juggler" class:

```as3
Starling.juggler.delayCall(myFunction, 1.5);
```

That will call "myFunction" in 1.5 seconds, with just one line of code. And the "DelayedCall" will be pooled for your, to avoid GC work. Were you aware of that? Does that help?

nomikovsupe:
Hi! Please forgot my previous example. And look at this example:


Now i can do what i want that way:
```javascript
private var mCall:Function;

private function someFunction():void
{
    mCall();
    mJuggler.repeatCall(myFunction, 10, 3);
}
```

 

But i prefer to do it that way:
```javascript
private var mDelayedCall:DelayedCall;

private function someFunction():void
{
    mDelayedCall.call();
    mDelayedCall.reset(myFunction, 10);
    mDelayedCall.repeatCount = 3;
    mJuggler.add(mDelayedCall);
}
```
 
It is look strange, that i want to write 4 lines instead of only 2. Yes. But i dont like to create duplicate of mCall variable, that has already in DelayedCall.

Thanks for you reply!

PrimaryFeather:
Sorry for the long silence on this issue; I was busy with other features and then forgot about it!
I've got another idea: what do you think of me adding a new method to the "DelayedCall" class that executes the callback with the given arguments right away?

```as3
private var mDelayedCall:DelayedCall;

private function someFunction():void
{
    mDelayedCall.execute();
    mDelayedCall.reset(myFunction, 10);
    mDelayedCall.repeatCount = 3;
    mJuggler.add(mDelayedCall);
}
```

The advantage: this would even take the given arguments into account. What do you think?

nomikovsupe:
Yes, it will be more appropriate than my approach. Thanks.

nomikovsupe:
Sorry but now i think that DelayedCall class would be dirty after this modification. DelayedCall would have unclear purpose.

PrimaryFeather:
Do you really think that? E.g. you might have a delayed call, but half into its lifetime you find out that the callback should be executed right away. Then you could do that and remove the delayed call from the juggler. 

We might also give it a different name, like "executeNow", "complete" or "completeNow".

nomikovsupe:
When i wrote my previous message, i thinked that delayedCall don't complete when we call DelayedCall.execute().

That's why i said that this behavior is dirty.


But if delayedCall will complete when we call DelayedCall.execute(), then all right. That case sounds good.


Sorry my english.

PrimaryFeather:
No need to apologize, your English is perfectly readable! :smile:
I called the method "complete", to show it's connection to the "isComplete" property. That should solve your problem, right? Otherwise, just let me know. =)

PR #677, opened 2015-02-15
Fixed possible memory leak when using async ATF texture upload.

Rokannon:
Unlike Starling objects native Flash textures do not remove their event listeners when texture is being disposed. Calling 'ConcreteTexture::dispose' right after 'ConcreteTexture::uploadAtfData' leaves Starling- and Flash- texture objects referencing each other which is increasing cost of garbage collection as these two objects can not be GC'ed during reference counting phase. This can even lead to a memory leak. Second possible problem related to this issue is async callback triggering on a disposed texture which can introduce unexpected behavior in application.


PrimaryFeather:
You're right, good catch! I merged your changes into the master branch.
Thanks a lot for the excellent fix!

PR #682, opened 2015-02-27
avoiding Texture size overflow when creating TextField with long long texts by lowering its resolution.

drumsoft:
This is a feature that avoid texture size exceeds the system limit when you create TextFields contains very large font or long string.
If bitmap data from the native flash TextField has over 2048 pixels width or height, It 'll be re-rendered lowering the resolution to fit in 2048x2048.

### DIFF
--- a/starling/src/starling/text/TextField.as
+++ b/starling/src/starling/text/TextField.as
@@ -90,6 +90,8 @@ package starling.text
     {
         // the name container with the registered bitmap fonts
         private static const BITMAP_FONT_DATA_NAME:String = "starling.display.TextField.BitmapFonts";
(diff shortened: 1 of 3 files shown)


PrimaryFeather:
Thanks a lot for the pull request! That's an excellent idea, and you implemented it very elegantly. Kudos! :smile:

Since "baselineExtended" and beyond support textures with a size of 4096x4096 pixels, I added support for that, as well. Furthermore, I noticed that we can dispose the bitmapData that's too big right away.

Again, thanks for your help!

drumsoft:
Thank you for merging and improve my patch!!
Texture.maxSize() is what I wanted when working on this problem :)

PR #683, opened 2015-02-27
fixed: AssetManager fails loading when enqueued flash.filesystem.File contains Unicode non-ascii path string.

drumsoft:
AssetManager fails loading when enqueued flash.filesystem.File contains Unicode non-ascii path string.
This patch fixes it.
Additionally, I added AIR version of test suite to certify it.


PrimaryFeather:
Good catch, thanks a lot for the pull request!

I merged the AssetManager fix right away; the new unit test project will have to wait a little longer. It makes a lot of sense to add that, but perhaps I should consider making the unit test AIR-only ... that way, we could test all features and still have only one test.

In any case, thanks! :smile:

PrimaryFeather:
(Labeled 'Feature' for the unit-test part, since the unicode-part is already fixed. :wink:)

drumsoft:
Thanks!

PrimaryFeather:
I'll close this issue for now, since it's not realistic that the 1.x branch of Starling will see those new unit tests. When I update the unit tests, I'll create a separate issue.

drumsoft:
OK. Thank you :)

PR #687, opened 2015-03-04
Define Starling root after constructor

joshtynjala:
It would be helpful if we could fully initialize Starling before providing it with the root display object. Sometimes, I want to do something between Starling initializing and the root display object being created. Especially if I need multiple NativeWindows in AIR desktop.

There is precedence to creating a stage without a root in the native display list. In AIR, when you create a NativeWindow, it has a stage, but no root. If I were to wrap NativeWindow in a Feathers component, it would make sense to offer the same experience.

In my plans to support MXML in Feathers, I'd like to have a main application class that abstracts Starling initialization. It would be ideal if I could set up a Feathers theme before the root is created so that the root can be a Feathers component. It's technically possible now, but it requires manual work for developers that I would  rather abstract away.

In my implementation, I made the following changes:

1) The rootClass parameter in the constructor may be set to null. initializeRoot() checks for this new possibility.

2) I added a root setter that can accept a display object. If a root already exists, it will be removed and disposed. If Starling hasn't fully initialized, the new root will be saved until initializeRoot() is called. In initializeRoot(), the saved root will it be added to the stage, and Event.ROOT_CREATED will be dispatched. If the root setter is used after initializeRoot() has been called, the new root is immediately added to the stage without an event.

The normal case of passing a valid rootClass to the constructor should behave exactly the same.

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -250,7 +250,6 @@ package starling.core
                                  renderMode:String="auto", profile:Object="baselineConstrained")
         {
             if (stage == null) throw new ArgumentError("Stage must not be null");
-            if (rootClass == null) throw new ArgumentError("Root class must not be null");
             if (viewPort == null) viewPort = new Rectangle(0, 0, stage.stageWidth, stage.stageHeight);
             if (stage3D == null) stage3D = stage.stage3Ds[0];
 
@@ -448,12 +447,15 @@ package starling.core
         
         private function initializeRoot():void
         {
-            if (mRoot == null)
+            if (mRootClass != null && mRoot == null)
             {
-                mRoot = new mRootClass() as DisplayObject;
+                root = new mRootClass() as DisplayObject;
                 if (mRoot == null) throw new Error("Invalid root class: " + mRootClass);
-                mStage.addChildAt(mRoot, 0);
-            
+                dispatchEventWith(starling.events.Event.ROOT_CREATED, false, mRoot);
+            }
+            else if (mRoot != null && mRoot.parent == null)
+            {
+                mStage.addChildAt(root, 0);
                 dispatchEventWith(starling.events.Event.ROOT_CREATED, false, mRoot);
             }
         }
@@ -1020,6 +1022,12 @@ package starling.core
         /** The instance of the root class provided in the constructor. Available as soon as 
          *  the event 'ROOT_CREATED' has been dispatched. */
         public function get root():DisplayObject { return mRoot; }
+        public function set root(value:DisplayObject):void
+        {
+            if (mRoot != null) mRoot.removeFromParent(true);
+            mRoot = value;
+            if (mRoot != null && mContext != null) mStage.addChildAt(mRoot, 0);
+        }
         
         /** Indicates if the Context3D render calls are managed externally to Starling, 
          *  to allow other frameworks to share the Stage3D instance. @default false */


PrimaryFeather:
That's seems very reasonable — thanks for the pull request, Josh!
I'll give the other code a quick review to make sure that exchanging "root" doesn't cause any problems somewhere else, and then I'll merge that.

PrimaryFeather:
I thought about it a little more, and there might be one problem with that approach, causing it not to work the way you intended.

You say:
>  It would be ideal if I could set up a Feathers theme before the root is created so that the root can be a Feathers component.

You won't be able to set up a Feathers theme before Starling has initialized, because you can't create any textures before there's a valid Context3D. So you still need to wait for Starling having finished initialization.

However, with your code, when you pass "null" as root class to the Starling constructor, you won't be able to know when Starling has finished initialization (lacking the "ROOT_CREATED" event), so you won't know when you're able to initialize your root object.

(After all, this was the reason that the Starling constructor requires a class, not a display object instance: to have control over when that class is instantiated.)

What about the following alternative:
* You may pass `null` to the `rootClass` parameter of the Starling constructor
* When Starling is ready with initialization, and `mRootClass` is still null, it simply doesn't instantiate a root
* There's a `rootClass` setter. 
  * If you assign a class before Starling is ready, any assigned class will be used as part of the existing `initializeRoot()` logic.
  * If you assign it later, it will cause the root class to be instantiated and will dispatch "ROOT_CREATED" — because that's what it just did.

Would that work, too? Or is this still too restrictive?

PrimaryFeather:
Like that:

```as3
private function initializeRoot():void
{
    if (mRoot == null && mRootClass != null)
    {
        mRoot = new mRootClass() as DisplayObject;
        if (mRoot == null) throw new Error("Invalid root class: " + mRootClass);
        mStage.addChildAt(mRoot, 0);
        dispatchEventWith(starling.events.Event.ROOT_CREATED, false, mRoot);
    }
}

public function get rootClass():Class { return mRootClass; }
public function set rootClass(value:Class):void
{
    if (mRootClass && mRoot)
        throw new Error("Root class may not change after root has been instantiated");
    else if (mRootClass == null)
    {
        mRootClass = value;
        initializeRoot();
    }
}
```

joshtynjala:
> you won't be able to know when Starling has finished initialization (lacking the "ROOT_CREATED" event), so you won't know when you're able to initialize your root object.

Isn't it possible to listen for Event.CONTEXT3D_CREATE? I could initialize the theme in the event listener, and then set the root, like this:

	private function starling_context3DCreateHandler(event:Event):void
	{
		starling.removeEventListener(Event.CONTEXT3D_CREATE, starling_context3DCreateHandler);
		new MetalWorksMobileTheme();
		starling.root = new RootClass();
	}

Regardless, don't mind using a rootClass property to pass in a class instead. That's perfectly acceptable. My code would probably be mostly the same:

	private function starling_context3DCreateHandler(event:Event):void
	{
		starling.removeEventListener(Event.CONTEXT3D_CREATE, starling_context3DCreateHandler);
		new MetalWorksMobileTheme();
		starling.rootClass = RootClass;
	}

PrimaryFeather:
> Isn't it possible to listen for Event.CONTEXT3D_CREATE?

Yes, you could use that — though technically, it doesn't promise you anything about the other Starling components being ready. But it would work. :wink: 

In any case, since you're fine with the "rootClass" property, I'll add it like that! I think it's a better fit with the constructor logic; otherwise, I fear that developers would use the property as a workaround for not having to pass the class to the constructor, and end up surprised that they get a "context missing" exception. ;-)

I'll make the change right away!

PrimaryFeather:
I think that should do it! Please try it out and let me know if it works for you. Cheers!

neolit123:
>
In any case, since you're fine with the "rootClass" property, I'll add it like that! I think it's a better fit with the constructor logic; otherwise, I fear that developers would use the property as a workaround for not having to pass the class to the constructor, and end up surprised that they get a "context missing" exception. ;-)

it would have been an interesting addition, and i wouldn't have to do some of the gymnastics i currently do to instantiate Starling from the Starling.root class itself :-) (or at least adding a custom child to the Starling stage and using a dummy root)

hmm, well Class extends Object, so i would have done it in the following way:
(not sure you will like the idea)

new Starling class constructor:
```
Starling(rootPrototype:Object, stage:Stage, viewPort:Rectangle = null, stage3D:Stage3D = null, renderMode:String = auto, profile:Object = baselineConstrained)
```

^ first argument is now an Object!

if "rootPrototype" is a Function it's a callback like so:

```
var starling:Starling = new Starling(readyCallback, ...);
...
private function readyCallback(_starling:Starling):void
{
    _starling.root = new myRoot(myArgs);
    // implies CONTEXT3D_CREATE
    // flexibility!
    // also things like this will work now!:
    // _starling.root = this;
    //
    // at that point i would probably lock the setter for Starling.root() not to allow further changes!
}

```
else if the "rootPrototype" is a Class then that's the Starling root class to be created.

else if "rootPrototype" is something else an Error is thrown!

what do you think?

PrimaryFeather:
Hm, that's not a bad idea either, thanks for the suggestion!
May I ask, though, why you want/need to change the root class? I honestly didn't think there would be much demand for this, so I'd be interested to know the reason that you'd like more flexibility in this respect. Thanks a lot in advance, Lubomir!

joshtynjala:
> I think that should do it! Please try it out and let me know if it works for you. Cheers!

Works nicely! Thanks, Daniel.

neolit123:
>Hm, that's not a bad idea either, thanks for the suggestion!
May I ask, though, why you want/need to change the root class? I honestly didn't think there would be much demand for this, so I'd be interested to know the reason that you'd like more flexibility in this respect. Thanks a lot in advance, Lubomir!

here is a concept of a core-framework that wraps the underling rendering (Flash vs Starling).
by abstracting the UI assets part switching between Starling and Flash is a matter of a true / false toggle...ideally.

Main - document class, setup level class for the user, extends MainBase,
MainBase - accepts a list of ProjectDisplay instances via the Main's super() call, extends flash.display.Sprite
ProjectDisplay - entry level class for the user, extends ProjectDisplayBase, passes arguments to ProjectDisplayBase via super().
ProjectDisplayBase - instantiates Starling, handles UI, keyboard events, fonts etc, extends starling.display.Sprite
StarlingRoot - demanded by Starling as a Class in the constructor

once StarlingRoot is ADDED_TO_STAGE it's no longer needed as the user interacts with ProjectDisplay. ProjectDisplayBase removes StarlingRoot from the Starling stage and adds itself at(0). but this could break depending on the future internal uses of Starling.root.

the proposal can eliminate the need for StarlingRoot, as ProjectDisplay can become the Starling root.

to summarize, it enables:
- separation between the CONTEXT3D_CREATE and the Root creation steps from the user perspective
- the user can pass arguments to his root class (important when sub-classing) and not using "public static"
- the root class can be the class in which Starling is instantiated (.root = this)

eventually, i can send a PR + add docs / wiki updates etc, if you think that this can be developed further. i'm surprised that Josh was the first one to bring this up as it was one of the first things i've noticed when i started using Starling.

P.S.: the new Shape / Graphics additions are especially nice!

PR #708, opened 2015-05-05
Fixed TravisCI build

vpmedia:
Hi,

I've removed the homebrew-binary/adobe-air-sdk dependency from the build script and switched to direct install. This way it can not be broken so easily when the package's version is updated.

Best regards!
Andrew


PrimaryFeather:
Ah, that sounds like a good plan! Thanks a lot, Andrew!

PR #709, opened 2015-05-05
Add feature to gets DO's children recursively (optional).

LordXaoca:
Example: ```parent.getChildByName("child1.child2.child3", true);```

### DIFF
--- a/starling/src/starling/display/DisplayObjectContainer.as
+++ b/starling/src/starling/display/DisplayObjectContainer.as
@@ -209,12 +209,24 @@ package starling.display
                 throw new RangeError("Invalid child index");
         }
         
-        /** Returns a child object with a certain name (non-recursively). */
-        public function getChildByName(name:String):DisplayObject
+        /** Returns a child object with a certain name. There is option to search
+         * recursively. */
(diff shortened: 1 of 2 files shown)


PrimaryFeather:
I'm not entirely sure about this feature — is this really necessary? After all, this could also be done via a custom utility method.

Any other opinions?

(Thanks anyway for the pull request, in any case! I appreciate the suggestion!)

tconkling:
I will always advocate for unnecessary features to be left to extensions :)

But also, if this were part of Starling mainline, it doesn't make sense to
parse the child names from a string - the parsing step probably makes it
slower than just passing in a varargs array of children names, and the name
separator is application-specific. If you're using a '.' in your child
names, you'd need to use a / or whatever.

LordXaoca:
Hi, sorry for late answer. 
Yep, it is not necessary, but it is conveniently. I agreed that separator may be pulled out as argument, but on other hand we will get too many arguments, what not so good from API point.
About performane, if you don't want to search in children, performance will be the same, but if you wanna search, you will write the same code, so, imho, there is no difference.

PR #712, opened 2015-05-09
Fix documentation on max quads

girasquid:
I'm not sure if this should be 16383 or 16382 (16382 would be double the old value), but this updates the comment to match the constant.


PrimaryFeather:
Good catch, thanks! =)
(It's 16383, that's correct!)

PR #720, opened 2015-06-01
Added check in contextValid() method for empty context.driverInfo String

vpmedia:
Hi!

I've noticed in some exception log files that the Context3D.driverInfo string can be empty, and in this case the context is also invalid. (The Context3D.profile has null value both when the string is empty "''" or 'Disposed'). The problem occurs mostly using Flash Player 16.X.

Just for note: the documentation about the possible values is not really detailed: http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display3D/Context3D.html#driverInfo

Best regards!
Andrew

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -1102,7 +1102,14 @@ package starling.core
          *  internal code Starling can't avoid), so do not call this method too often. */
         public function get contextValid():Boolean
         {
-            return mContext && mContext.driverInfo != "Disposed";
+            if(mContext) {
+                const driverInfo:String = mContext.driverInfo;
+                if(driverInfo && driverInfo != "" && driverInfo != "Disposed") 
+                {
+                    return true;
+                }
+            }
+            return false;
         }
 
         // static properties


vpmedia:
The exception which I'm trying to solve..:

	Error: Error #3694: The object was disposed by an earlier call of dispose() on it.
	at flash.display3D::VertexBuffer3D/uploadFromVector()
	at starling.display::QuadBatch/syncBuffers()[starling\display\QuadBatch.as:210]
	at starling.display::QuadBatch/renderCustom()[starling\display\QuadBatch.as:222]
	at starling.core::RenderSupport/finishQuadBatch()[starling\core\RenderSupport.as:596]
	at starling.core::RenderSupport/applyClipRect()[starling\core\RenderSupport.as:423]
	at starling.core::RenderSupport/setRenderTarget()[starling\core\RenderSupport.as:372]
	at starling.core::RenderSupport/set renderTarget()[starling\core\RenderSupport.as:361]
	at starling.core::Starling/render()[starling\core\Starling.as:518]
	at starling.core::Starling/nextFrame()[starling\core\Starling.as:478]
	at starling.core::Starling/onEnterFrame()[starling\core\Starling.as:699]

Maybe connected to https://github.com/Gamua/Starling-Framework/pull/627 ?

PrimaryFeather:
Thanks, Andrew, that's interesting!

Did you literally see in your logs that the `driveInfo` string is empty, or do you just expect it to be empty, because it would explain the error you're getting?

In any case, it wouldn't surprise me if Adobe's engineers forgot to set the correct string in some situations.

vpmedia:
Yes, I'm logging the driverInfo string to php and it's an empty string (NOT null)

PrimaryFeather:
Thanks for the information, I was not aware of that!
I added that change right away. 
(With a slightly modified source style, but the logic is the same.)

vpmedia:
Thanks, I'll try to re-test the issue with my app (Losing context while uploading textures).

Oldes:
https://github.com/Oldes/Starling-Framework/commit/684439fe4c7fe1a1a2271b4d3b70cecc8bc766d0#commitcomment-11517922

PR #733, opened 2015-07-02
Avoid Rectangle object creation

Fraggle:
Just a quick fix, I guess it's harmless.


PrimaryFeather:
Jeff is right, the second one is too dangerous. If it's done that way, the static rectangle is passed outside the class, which means developers could end up storing it, and will be surprised that it constantly changes its value.

The first one is a good idea, though — totally overlooked that. Thanks!

PR #736, opened 2015-07-11
adding feature "frameEvents" to starling movieClip

dawsonmv:
I have added the ability to have an event that is dispatched on a specific frame of a starling.display.MovieClip.  Ideally this would replace frameSound because it is more ubiquitous.  In this version frameSound has been left intact and the functions getFrameEvent & setFrameEvent have been added as well as mEvents vector and dispatchFrameEvent function. The dispatchFrameEvent function is called after the playSound function is called. A new const was also added starling.events.Event.MOVIE_CLIP_FRAME_EVENT


dawsonmv:
once again, i have not here, but removing the sound dependancies, vars, and funcs from the movieClip class would make it more light weight. in addition frameEvents can be used for more things other than sound. In this implementation the frameEvents are stored as strings, these eventNames are attached as data to an Event.MOVIE_CLIP_FRAME_EVENT when dispatched. a listener can be used to play a sound or whatever is required by the developer for that frame.

PrimaryFeather:
Thanks a lot for the pull request! You're right, such a frame event makes a lot of sense, and is more useful than just the list of sounds. Actually, I planned to add that a while ago, but never found the time to look into it properly.

I can't merge that right now, though — I'll need time to take a closer look, and I'd maybe want to merge it all into one clean commit. But it's noted and will be looked into! 

Thanks again! I hope you can live with the delay.

dawsonmv:
i can open another pull request if you prefer that is a single commit. with even yet another pull request that has the sound bits removed. Might have some time later this evening. Once the sound bits are removed then there will be no confusion as to which events.Event class is being referenced.

PrimaryFeather:
That would help, yes! 

However, as an alternative, you could simply post a link (e.g. to a Gist) containing your updated MovieClip class. I will review it manually, anyway — and for that, such a link would suffice. That's really all I'd need.

Thanks in advance!

dawsonmv:
here is the gist, these use the old vector.splice and not the new flash19 vector methods:

Starling MovieClip with frameEvents with sound features intact:
https://gist.github.com/3efbd8e9aff150c9686a.git

Starling MovieClip with frameEvents and sound feature removed:
https://gist.github.com/6707ef1d935d8324d280.git

PrimaryFeather:
Thanks a lot for that!
I just reopened the pull request, so that it isn't overlooked.

PrimaryFeather:
Actually, I integrated this in Starling 2, heavily inspired by your pull request. I just used functions instead of events, and I kept sound handling as it is (since I could write it in a slightly optimized way like that -- sounds can't have any side effects like "stop movie").

I totally forgot to close the pull request -- doing so now.
Thanks again for your help, and sorry that it took so long to get this into the core.

PR #738, opened 2015-07-14
Merge pull request #1 from Gamua/master

samao:
update my fork


PrimaryFeather:
Hi Samao!

I hope this doesn't sound like nit-picking, but this is still not optimal. If you look at the pull request, you'll see that it didn't start with the latest Starling version. Thus, it would revert some of the changes that I recently made.

However, as an alternative, you could simply post a link (e.g. to a Gist) containing your updated MovieClip class. I will review it manually, anyway — and for that, such a link would suffice. 

Either way: thanks for your understanding!

PrimaryFeather:
I just noticed that I mistake your pull request for another one — sorry for the misleading post above (which I already deleted).

However, I still don't get what this request is all about? There is no information about what you're actually trying to achieve. Perhaps you could elaborate on that?

PR #754, opened 2015-08-13
Add "requiresRedraw" to starling.text.TextField

MuaoLoniel:
Please add "requiresRedraw" to starling.text.TextField

I need it in my class:
```
public class WordWrapTextField extends starling.text.TextField
{
	private var mWordWrap:Boolean;
 
	override protected function formatText(textField:flash.text.TextField, textFormat:TextFormat):void
	{
		super.formatText(textField, textFormat);
		textField.wordWrap = mWordWrap;
	}
	 
	public function get wordWrap ():Boolean { return mWordWrap }
	public function set wordWrap (value:Boolean):void
	{
		mWordWrap = value;
		super.requiresRedraw();
	}
}
```
Thanks.

### DIFF
--- a/starling/src/starling/text/TextField.as
+++ b/starling/src/starling/text/TextField.as
@@ -253,6 +253,11 @@ package starling.text
          *  @param textFormat the default text format that's currently set on the text field.
          */
         protected function formatText(textField:flash.text.TextField, textFormat:TextFormat):void {}
+        
+        final protected function requiresRedraw():void 
+        {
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Thanks, that makes sense!

I just renamed the method to "requireRedraw" (note the missing "s"), though. The third person "s" makes sense for the member variable, but the method should be a verb. 

... yeah, I'm a nit-picker, I know. :grin:

Thanks anyway!

PR #760, opened 2015-08-30
Fix issue with Textfield when native filters are applied

MuaoLoniel:
```
// calculateFilterOffset bug
var tf:TextField = new TextField(100, 40, 'Playgame', "Verdana", 15);
tf.hAlign = HAlign.RIGHT;
tf.border = true;
tf.nativeFilters = [new DropShadowFilter(20, 0, 0xFF00F0, 1, 1, 1, 1)]
addChild(tf);
tf.x = 50;
tf.y = 20;

// autosize bug
var tf2:TextField = new TextField(100, 40, 'Playgame', "Verdana", 15);
tf2.autoSize = TextFieldAutoSize.HORIZONTAL;
tf2.border = true;
tf2.nativeFilters = [new DropShadowFilter(20, 180, 0xFF00F0, 1, 1, 1, 1)]
addChild(tf2);
tf2.x = 50;
tf2.y = 80;

// autoscale bug
var tf3:TextField = new TextField(100, 40, 'Playgame', "Verdana", 45);
tf3.autoScale = true;
tf3.border = true;
tf3.nativeFilters = [new DropShadowFilter(20, 180, 0xFF00F0, 1, 1, 1, 1)]
addChild(tf3);
tf3.x = 50;
tf3.y = 140;

```

untitled-1

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -587,7 +587,8 @@ package starling.core
                                              wantsBestResolution:Boolean=false):void
         {
             enableDepthAndStencil &&= SystemUtil.supportsDepthAndStencil;
-
+			width = (width < 32) ? 32 : width;
+			height = (height < 32) ? 32 : height;
             var configureBackBuffer:Function = mContext.configureBackBuffer;
             var methodArgs:Array = [width, height, antiAlias, enableDepthAndStencil];
             if (configureBackBuffer.length > 4) methodArgs.push(wantsBestResolution);
--- a/starling/src/starling/text/TextField.as
+++ b/starling/src/starling/text/TextField.as
@@ -260,7 +260,9 @@ package starling.text
         {
         	mRequiresRedraw = true;
         }
-
+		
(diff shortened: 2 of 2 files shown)


PrimaryFeather:
Thanks a lot for the fixes! I'll review the changes as soon as I've got some time. =)

PrimaryFeather:
Just a quick heads-up: I'll probably look at this a little later this year, when I'm revisiting the TextField class for Starling 2.0. I plan some other changes in that respect, as well, and that way I can tackle it all at once.

PrimaryFeather:
I decided to remove the `nativeFilters` property completely in Starling 2. I know not everybody will be happy about this, but I decided that the time would be better spent working on Starling's filter API. And the good news is that Starling filters are now so powerful that we shouldn't need the old filters any longer. Yes, not every filter is there yet, but they can be created on a case-by-case basis; and performance-wise, that will be even faster than the (CPU-created) classic filters.

I hope you can understand my reasons!

PR #763, opened 2015-09-11
Allow setting the pma setting for an ATF texture (closes #762)

### LINKED ISSUE #762: Blending of PMA with non PMA
Hello Daniel,

We were having issues when we used some blend modes, especially additive with non PMA textures from ATF.

Example:
atfstandard
The background on the left side is a PNG (so PMA) and on the right side an ATF (so non PMA) with blend mode none.
The 2 green spots at the top are PNG (so PMA) in additive.
The 2 green spots at the bottom are ATF (so non PMA) in additive.
There are also a white cloud PNG between the two top spots and a white cloud ATF between the two bottom spots in additive.
In the middle there is a render texture with only the alpha channel.

You can see that there is an issue where the alpha is not zero for the ATFs.
It also appear when blendMode is Normal for non PMA on top of PMA.

We found out that the issue happen because the blend mode factors only take into account the source PMA state and expect the destination to be of the same PMA state. As we cannot know the destination PMA state (it can even vary from pixel to pixel), we came up with 2 solutions that would fix the bug.

The first solution is already discussed in another issue, we could alter the quadbatch fragment shader to alter the output color when not PMA to premultiply everything or the other way around, just to make sure everything is the same. Of course this is not without a potential performance impact since we add a MUL or a DIV in the shader for every pixel drawn from certain texture.

The other solution is simply to premultiply the source image supplied to the ATF converter and allow us to say wether the ATF is PMA or not (as right now you force the PMA flag to off when uploading an ATF). That way everything would be PMA and it will works flawlessly.

So my request would be for you to add a pma flag to TextureOptions and an additional argument to fromAtfData(..) plus the additional plumbing between AssetManager and Texture. It is transparent for existing users and it would fix all our issues with PMA / non PMA.

Here is the same example where we did just that : manually premultiplied the png before submitting to png2atf then forced the pma to true in fromAtfData().
atfwithpma

I can do the patch myself and submit it if you want.
What do you think ?

Seb

Fraggle:
Here goes the pull request.

### DIFF
--- a/starling/src/starling/textures/Texture.as
+++ b/starling/src/starling/textures/Texture.as
@@ -161,7 +161,8 @@ package starling.textures
             else if (data is ByteArray)
             {
                 texture = fromAtfData(data as ByteArray,
-                    options.scale, options.mipMapping, options.onReady, options.repeat);
+                    options.scale, options.mipMapping, options.onReady, options.repeat,
+					options.premultipliedAlpha);
             }
             else
                 throw new ArgumentError("Unsupported 'data' type: " + getQualifiedClassName(data));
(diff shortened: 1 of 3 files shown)


Fraggle:
It does not work reliably with DXT check #762 .

PR #764, opened 2015-09-14
Convert to PMA at the shader if needed and wanted (#762)

Fraggle:
Hi Daniel,

I tried to be a little invasive as possible and to follow your "style".
I added a flag on starling to request the non PMA texture to be converted to PMA, false by default so it should keep the old code exactly as fast as before.
If the flag is set to true, then It'll convert the non PMA texture to PMA in the shader and force blending to consider everything as PMA (regardless of the original texture format).

It fix all our blending issues.

Thanks for your time (Simas told me that v2-beta is across the corner, can't wait!)

Seb

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -39,7 +39,7 @@ package starling.core
     import flash.utils.Dictionary;
     import flash.utils.getTimer;
     import flash.utils.setTimeout;
-
+    
     import starling.animation.Juggler;
     import starling.display.DisplayObject;
     import starling.display.Stage;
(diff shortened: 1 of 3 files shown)


PrimaryFeather:
Thanks a lot, that's a very clean implementation! I'm currently deep into the 2.0 version, but as soon as it's possible, I'll have another look and will probably merge this.

PrimaryFeather:
I just realized that I didn't give you any more feedback on this; but you'll probably already have seen that I implemented it like that in Starling 2! 

In fact, I decided to not even make this optional, but force this behavior. It was your pull request that finally pushed (pulled?) me to doing that, and this is a change that I'm extremely happy with. It made a lot of things MUCH easier.

Thanks again! I'll close this issue, then. :smile:

PR #770, opened 2015-09-23
fix travis

honzabrecka:
by updating player version

PR #771, opened 2015-09-23
travis: use specific version of AIR SDK

honzabrecka:
Use specific version of AIR SDK to prevent incompatibility issues (related to #770).


PrimaryFeather:
Aaaaah, that explains! Thanks a lot, that definitely explains the problems during the last days. :wink:

Issue #778, opened 2015-10-25
FragmentFilter passTexture is always a RectangleTexture


devon-o:
In the FragmentFilter.updatePassTextures() method, the pass textures are created with the call: Texture.empty(width, height, PMA, false, true, scale). This means that unless using the 'baselineConstrained' profile, a RectangleTexture is always created. While this is nice performance-wise, a RectangleTexture cannot be wrapped/repeated - so if the AGAL shader in the filter being applied uses the 'repeat' or 'wrap' flag in the texture sample, the filter will fail. I haven't found a clean solution to suggest, but it seems the pass textures should be created with the repeat property set to the repeat property of the filtered display object's texture.

neolit123:
hello Devon,

i think you have a point about the optional repeat, but a FragmentFilter target DisplayObject can also be a Quad, which doesn't have a Texture, so it can't really be obtained reliably from any target DisplayObject. also the target can be a deep tree of containers and a mixture of Texture DisplayObjects which have different texture settings.

it might be a good idea to simply expose the "repeat" (and/or "mipMapping") argument(s) which is (are) passed to Texture.empty() in FragmentFilter.updatePassTextures() as getter/setters in FragmentFilter. setting at least the optional "repeat" to "true" would not create a RectangleTexture for Context3D in Texture.empty().

if you modify FragmentFilter in such a way, does it work for your test case?
not sure how that would resonate as a whole...

devon-o:
Good point about the quads and nested DO's. Adding a protected boolean 'repeat' property to the FragmentFilter (in a similar way as 'mode', 'numPasses', etc) does the trick. You can just set 'repeat' to true in your own FragmentFilter implementation and everything is fine. I may do a pull request later after I take another look with fresher eyes.

d.

PR #779, opened 2015-10-26
Added repeat property to FragmentFilter to allow wrapping AGAL shaders.

devon-o:
This is a fix for issue 778 (https://github.com/Gamua/Starling-Framework/issues/778)

### DIFF
--- a/starling/src/starling/filters/FragmentFilter.as
+++ b/starling/src/starling/filters/FragmentFilter.as
@@ -91,6 +91,7 @@ package starling.filters
         
         private var mNumPasses:int;
         private var mPassTextures:Vector.<Texture>;
+        private var mRepeat:Boolean;
 
         private var mMode:String;
         private var mResolution:Number;
@@ -388,7 +389,7 @@ package starling.filters
                 disposePassTextures();
 
                 for (var i:int=0; i<numPassTextures; ++i)
-                    mPassTextures[i] = Texture.empty(width, height, PMA, false, true, scale);
+                    mPassTextures[i] = Texture.empty(width, height, PMA, false, true, scale, "bgra", mRepeat);
             }
         }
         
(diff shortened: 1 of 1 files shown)


neolit123:
so as the Travis CI build indicates, there seems to be a conflict with the "public get/set repeat()" in DisplacementMapFilter(). i have overlooked that...

"protected function get/set textureRepeat()" paired with "mTextureRepeat" in FragmentFilter might be the non-breaking solution.

devon-o:
Interesting - I overlooked that as well. A 'textureRepeat' property on FragmentFilter may be a quick fix, but DisplacementMapFilter also needs some work as its 'repeat' property is currently broken due to this same issue.

PrimaryFeather:
Thanks a lot for the pull request, and for opening up the issue!

The reason I defaulted to `repeat = false` is that I didn't see any use-case where you'd be able to make use of a wrapping pass texture. As @neolit123 mentioned, the filtered object may have a size of, say, 129x129 pixels, which would create a pass texture with 256x256 pixels (they are always created in POT-style, I think), with only the top left quarter being used. How could you make use of any wrapping with that limitation? 

On the other hand, I probably ran into exactly the same problem with the DisplacementMapFilter, since I added the property there, too. Even though (as you noted) that seems to be broken, :grin:.

So I'd add that property to the FragmentFilter base class and use it in the DisplacementMapFilter, too. Would that work for you?

devon-o:
Hi Daniel,

I considered doing just that (and it would work for my use cases), but ran into another problem. Because the repeat property in DisplacementMapFilter is public, it would be perfectly legit (albeit odd) for a user to change the repeat property from false to true after creating and applying the filter. This is particularly nasty as the mPassTextures vector in FragmentFilter has been filled, so the pass textures won't be (re)created and the filter is left in a broken state at that point. I considered making the repeat property of DisplacementMapFilter protected (or inheriting from the protected FragmentFilter.repeat property I was adding), but that's a potentially app breaking api change for folks currently using the DMFilter. So I left it for now, but I'm open to suggestions (or can give more details in case that didn't make sense).

PR #787, opened 2015-11-23
Make Button.MAX_DRAG_DIST public and class variable

GRIM2D:
It makes easy customizing button behaviour.


PrimaryFeather:
Thanks a lot for the pull request! You're right, it makes sense to expose that setting. Although I'll probably do it as a property; right now, with the uppercase style, it's breaking the coding conventions I'm typically using.

I'll look into it!

PR #797, opened 2016-02-04
DisplayObjectContainer now uses linkedlist to store children

phendrax:
faster addChild/removeChild and faster looping through children

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -364,7 +364,7 @@ package starling.core
             {
                 _root = new _rootClass() as DisplayObject;
                 if (_root == null) throw new Error("Invalid root class: " + _rootClass);
-                _stage.addChildAt(_root, 0);
+                _stage.insertChildBeforeAll(_root);
 
                 dispatchEventWith(starling.events.Event.ROOT_CREATED, false, _root);
             }
--- a/starling/src/starling/display/DisplayObject.as
+++ b/starling/src/starling/display/DisplayObject.as
@@ -20,7 +20,7 @@ package starling.display
     import flash.ui.Mouse;
     import flash.ui.MouseCursor;
     import flash.utils.getQualifiedClassName;
-
+    
     import starling.core.Starling;
     import starling.errors.AbstractClassError;
     import starling.errors.AbstractMethodError;
@@ -129,6 +129,8 @@ package starling.display
     public class DisplayObject extends EventDispatcher
     {
         // private members
+		public var prev:DisplayObject;
+		public var next:DisplayObject;
         
         private var _x:Number;
         private var _y:Number;
@@ -509,16 +511,7 @@ package starling.display
         /** @private */
         internal function setParent(value:DisplayObjectContainer):void 
         {
-            // check for a recursion
-            var ancestor:DisplayObject = value;
-            while (ancestor != this && ancestor != null)
-                ancestor = ancestor._parent;
-            
-            if (ancestor == this)
-                throw new ArgumentError("An object cannot be added as a child to itself or one " +
-                                        "of its children (or children's children, etc.)");
-            else
-                _parent = value;
+            _parent = value;
         }
         
         /** @private */
--- a/starling/src/starling/display/DisplayObjectContainer.as
+++ b/starling/src/starling/display/DisplayObjectContainer.as
@@ -15,7 +15,7 @@ package starling.display
     import flash.geom.Rectangle;
     import flash.system.Capabilities;
     import flash.utils.getQualifiedClassName;
-
+    
     import starling.core.starling_internal;
     import starling.errors.AbstractClassError;
     import starling.events.Event;
@@ -66,7 +66,10 @@ package starling.display
     {
         // members
 
-        private var _children:Vector.<DisplayObject>;
+		public var childFirst:DisplayObject;
+		public var childLast:DisplayObject;
+        private var _numChildren:int = 0;
+		
         private var _touchGroup:Boolean;
         
         // helper objects
@@ -86,15 +89,17 @@ package starling.display
             {
                 throw new AbstractClassError();
             }
-            
-            _children = new <DisplayObject>[];
         }
         
         /** Disposes the resources of all children. */
         public override function dispose():void
         {
-            for (var i:int=_children.length-1; i>=0; --i)
-                _children[i].dispose();
+			var child:DisplayObject = this.childFirst;
+			var next:DisplayObject = child;
+           	while(next) {
+				next = child.next;
+				child.dispose();
+			}
             
             super.dispose();
         }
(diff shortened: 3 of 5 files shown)


neolit123:
hello,

1) the patch is not following the Startling coding style
2) for a big change like that you should provide benchmarks to support your claim "faster looping through children".

PrimaryFeather:
Thanks a lot for your efforts, phendrax!
However, as Lubomir correctly added, it would be great to have some data about the actual gain of this change.

I'm confident that iterating over the objects via a linked list is going to be faster than using a Vector; however, that object access is only a very tiny part of what's done in the render loop, so I'm not sure if it's worth the additional complexity. I'm a fan of keeping the code as simple as possible. :wink:

neolit123:
in the past i've seen benchmarks that AS3.0 Vectors, which are a native (e.g. C) implementation, are slower than an inbreed AS3.0 linked list which makes no sense, unless something is wrong in the Vector implementation or the VM reference caching. i've also seen similar odd results in Java related topics, so this could be a VM only phenomenon.

for AIR/Flash there might have been some issues in the past, because i just did my first test on this topic and i cannot confirm any performance benefits of a dual-linked-list over a Vector for my use case (see source code and results bellow). 

i think at some point someone dropped the suggestion that linked-list are better than Vectors in every aspect, for any programming language and they started trending. this is of course wrong. 

linked-lists have their uses in certain object oriented structures (e.g. if you want to write a XML parser), but really should be considered as the more nitsche storage type. same goes for the MVC design pattern - at some point someone suggested that MVC is the best solution to framework design and every single library started using it, which is of course a similarly bad call to always using linked-lists for data storage.

```
/*
a test with AIR 20 on Win32 about the performance of Vectors against linked
lists.

COMPILATION:
amxmlc -debug=false Main.as

RUNTIME:
adl Main-app.xml

RESULTS:
** number of elements to test with 2097152
* test random read access of 8192 elements
vector: time (ms) 0
linked list: time (ms) 49885
* test push of a single element
vector: time (ms) 0
linked list  (not keeping an end-pointer): time (ms) 10
* test random delete of ~8192 elements
vector: time (ms) 4381
linked list: time (ms) 49692
*/
package
{
	import flash.desktop.NativeApplication;
	import flash.display.Sprite;
	import flash.utils.getTimer;

	public class Main extends Sprite
	{
		private static const LEN:uint = 1 << 21;
		private static const LEN_ACCESS:uint = LEN >> 8;

		private const vector:Vector.<TestObject> = new Vector.<TestObject>(LEN, false);
		private const linkedList:LinkedList = new LinkedList(getRandValue());
		private var linkedListLen:uint = LEN;
		private const indexes:Vector.<uint> = new Vector.<uint>(LEN_ACCESS, true);

		public function Main():void
		{
			var i:uint, value:uint, idx:uint, len:uint;
			var timer:int;

			// fill a Vector
			for (i = 0; i < LEN; i++)
				vector[i] = new TestObject(getRandValue());

			// fill LinkedList
			var ptr:LinkedList = linkedList, ptrTmp:LinkedList;
			for (i = 0; i < LEN; i++) {
				ptrTmp = new LinkedList(getRandValue());
				ptr.next = ptrTmp;
				ptrTmp.prev = ptr;
				ptr = ptrTmp;
			}

			// fill a Vector of indexes to be accessed in the Vector and LinkedList
			for (i = 0; i < LEN_ACCESS; i++)
				indexes[i] = getRandValue();

			trace("** number of elements to test with", LEN);

			trace("* test random read access of " + LEN_ACCESS + " elements");
			len = LEN_ACCESS;
			// test random read access
			timer = getTimer();
			for (i = 0; i < len; i++)
				value = vector[indexes[i]].value;
			trace("vector: time (ms)", getTimer() - timer);

			timer = getTimer();
			for (i = 0; i < len; i++)
				value = linkedListAtIndex(indexes[i]).value;
			trace("linked list: time (ms)", getTimer() - timer);

			// test push
			trace("* test push of a single element");
			timer = getTimer();
			vector.push(new TestObject(getRandValue()));
			trace("vector: time (ms)", getTimer() - timer);

			timer = getTimer();
			linkedListPush(new LinkedList(getRandValue()));
			trace("linked list (not keeping an end-pointer): time (ms)", getTimer() - timer);

			// test random delete
			len = LEN_ACCESS;
			trace("* test random delete of ~" + len + " elements");
			timer = getTimer();
			for (i = 0; i < len; i++) {
				idx = indexes[i];
				if (idx < vector.length - 1)
					vector.removeAt(idx);
			}
			trace("vector: time (ms)", getTimer() - timer);

			timer = getTimer();
			for (i = 0; i < len; i++) {
				idx = indexes[i];
				if (idx < linkedListLen - 1)
					linkedListRemoveAt(idx);
			}
			trace("linked list: time (ms)", getTimer() - timer);

			// ADL / AIR 20 doesn't automatically exit for some reason
			NativeApplication.nativeApplication.exit();
		}

		private function getRandValue():uint
		{
			return Math.random() * LEN;
		}

		private function linkedListAtIndex(_idx:uint):LinkedList
		{
			var ptr:LinkedList = linkedList;
			var counter:uint = 0;
			while (ptr) {
				if (counter == _idx)
					return ptr;
				ptr = ptr.next;
				counter++;
			}
			return null;
		}

		private function linkedListPush(_new:LinkedList):void
		{
			var last:LinkedList = linkedListAtIndex(linkedListLen - 1);
			last.next = _new;
			_new.prev = last;
			linkedListLen++;
		}

		private function linkedListRemoveAt(_idx:uint):void
		{
			var ptr:LinkedList = linkedListAtIndex(_idx);
			var prev:LinkedList = ptr.prev, next:LinkedList = ptr.next;
			if (prev)
				prev.next = next;
			if (next)
				next.prev = prev;
			linkedListLen--;
		}
	}
}

final class TestObject
{
	public var value:uint;

	public function TestObject(_value:uint):void
	{
		value = _value;
	}
}

final class LinkedList
{
	public var prev:LinkedList = null;
	public var next:LinkedList = null;

	public var value:uint;

	public function LinkedList(_value:uint):void
	{
		value = _value;
	}
}
```

phendrax:
Hi,

thanks for you replies

I agree that this patch doesn't follow Starling's coding style (partly because it doesn't implement *ChildAt methods) but I thought I could share this since it makes Starling a little more faster

here are some benchmarks I did:

with Vector:
```
const LEN:int = 30000;
var t:int = getTimer();
for(var i:int=0; i<LEN; i++) {
  var s:Quad = new Quad(100,100);
  map.addChild(s);
}
trace("addChild", (getTimer()-t)+"ms");

t = getTimer();
const len:int = map.numChildren;
for(i=0; i<len; i++) {
  map.getChildAt(i);
}
trace("loop", (getTimer()-t)+"ms");

t = getTimer();
for(i=len-1; i>=0; i--) {
  map.removeChildAt(0); // waaaay faster to removeChildAt(0) than removeChildAt(i)
}
trace("remove", (getTimer()-t)+"ms");
```

with LinkedList:
```
const LEN:int = 30000;
var t:int = getTimer();
for(var i:int=0; i<LEN; i++) {
  var s:Quad = new Quad(100,100);
  map.addChild(s);
}
trace("addChild", (getTimer()-t)+"ms");

t = getTimer();
var child:DisplayObject = map.childFirst;
while(child) {
  child = child.next;
}
trace("loop", (getTimer()-t)+"ms");

t = getTimer();
child = map.childLast;
var prev:DisplayObject;
while(child) {
  prev = child.prev;
  map.removeChild(child);
  child = prev;
}
trace("remove", (getTimer()-t)+"ms");
```

Here are the results:

Vector:
```
addChild 3612ms
loop 24ms
remove 780ms
```

LinkedList:
```
addChild 2857ms
loop 2ms
remove 432ms
```

As you can see, looping is pretty much the same in both case (I think the LinkedList impl is only faster because we dont have a function call (getChildAt)).
The removeChild method would be even more slower with Vector if we removed the last child (because it calls indexOf internally)

Thanks for having had a look :)

b005t3r:
But you do realize random access is going to be a performance killer with linked lists, right?

phendrax:
Hi booster,
Yes of course, using getChildAt or removeChildAt would be very slow, but personally, I rarely (not to say never) use those methods outside a loop.
Morever, current removeChildAt internally uses vector.indexOf() which is (imo) a performance killer too

b005t3r:
Well, I do use it a lot, so I'm voting against this change :) Plus, the most often used way of removing children, which is removeFromParent(), wouldn't benefit from this change as well.

phendrax:
Hi,
removeFromParent() is just:
`if(this.parent) this.parent.removeChild(this);`
which is way more faster with a LinkedList than a Vector
(cause using a vector you have to first get the index of the child before remiving it)

neolit123:
you are correct about removeChild() - that's one of the things linked-lists are good at.

i see a bit of a problem in the Flash API. the (recently added) Vector.removeAt() call needs an index, and to get the index, Vector.indexOf() is used. if there is a Vector.remove(object:T) then things will be much faster, because there won't be any intermediate AS3.0 objects and a couple of calls, but a single native call.

yet, the linked-list removeChild() may still be faster because you simply manipulate a couple of references to remove this particular Object. linked-lists would also be very good at things like addAbove(), addBellow(), removeBellow(), removeAbove() and so on...also swapChildren() (by reference).

let's be objective here and say the two most used functions in the DOC class are addChild() / removeChild(). in that case linked-lists will be faster, because addChild() is pretty much the same for both constructs. the problem is that the moment the user introduces any of the *At() API the performance overall drops, and it drops a lot.

for N elements, indexed access is 50000 times slower for linked-lists, while i just did some tests for a removeByReference() (e.g. removeChild()) and Vectors are 5000 times slower, mostly due to the indexOf() culprit. without the indexOf() culprit explained above, let's estimate a 2000 times slower removeByReference() for a Vector.

sorting is something else the Starling DOC class supports via sortChildren() and linked-lists are also particularly bad at that. either the "not implemented" change in your patch should be reflected in the documentation or a merge-sort algorithm for the linked-list should implemented.

ideally, the whole current Starling DOC API should be implemented with the linked-list and broken down in a performance table to be compared with the Vector counterpart, possibly with a common-usage weighting factor (e.g. addChild() is more common) and with some memory consumption estimates. i will be surprised if the linked-list "outweights" the Vector.

PR #800, opened 2016-02-12
fix travis ci

honzabrecka:
I just copied the .travis.yml from master branch. See #771 for more info.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@
 #
 # =================================================================================================
 
-git: 
+git:
     depth: 10
 
 language: objective-c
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Ah, damn ... I thought that change was before I introduced the v2.0 branch ... :wink:
Thanks a lot for your help!

PrimaryFeather:
Sorry to bug you again, Honza. ;-) 

For whatever reason, the test now never finishes, but I can't see any indication why that would happen. I recently added a new test class (FilterChainTest), I don't know if that's connected ... 

Do you have any idea?

PR #801, opened 2016-02-15
Fix gradle build

honzabrecka:
Test task hanged because of `ReferenceError: Error #1069: Property profile not found on flash.display3D.Context3D and there is no default value.` This property was added in some newer version of FP, but because the build was using 11.1, it wasn't available.

Cheers!


PrimaryFeather:
Fantastic! AIR 19 is the minimum required version for Starling 2, so you picked just the right reference, too. Thanks a lot!

May I ask where you saw that it was that error? I couldn't find any information other than the build was hanging ...

honzabrecka:
I run gradle localy the same way as the travis does. :)

PrimaryFeather:
Ha, okay, that sounds logical. :wink:
Thanks again!

PR #816, opened 2016-03-22
[~] Added a returnPlay property to MovieClip.

Binklebonk:
This is a feature I've put into a current project.  When enabled, the clip will play to the end, then effectively play in reverse back to the start, in a back-forth way.  I thought I'd see if it could be worked into the MovieClip class as a built in property.

In effect, it replicates the existing frames and adds them in reverse to the MovieClipFrames vector.

In handling things like numFrames and currentFrame, I've obscured the return value to only reflect the original clip.  The reasoning is I thought it'd be easier for people to continue treating the clip as unchanged in length and framecount, it simply plays in reverse when it gets to the end.

I'm uncertain if currentTime should be obscured in the same way, my gut feeling is it should remain unobscured, so the full play time back and forth, is twice the normal play time.

When adding and removing frames to a clip, if returnPlay is active, the logic works like this:
if (returnPlay == true)
returnPlay = false: remove the mirrored frames
add or remove frames
returnPlay = true: re-mirror the frames

There's perhaps a more elegant way to do this rather than destroying and re-creating all the mirrored frames each time, but it's not something I anticipate needs to be done too frequently so I've left it for now.

Known quirks: The first and last frames of the clip are doubled.  I'm not sure if this is an issue, or expected behaviour?

### DIFF
--- a/starling/src/starling/display/MovieClip.as
+++ b/starling/src/starling/display/MovieClip.as
@@ -49,6 +49,7 @@ package starling.display
         private var _currentTime:Number;
         private var _currentFrameID:int;
         private var _loop:Boolean;
+		private var _returnPlay:Boolean;
         private var _playing:Boolean;
         private var _muted:Boolean;
         private var _wasStopped:Boolean;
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
I'm currently working at bug fixes and performance improvements of Starling 2, but as soon as that's done, I'll have a detailed look at your proposal. Thanks for your efforts! :smile:

Binklebonk:
No worries Daniel, all that you're doing in Starling 2 is hugely
appreciated 😀

David

PR #825, opened 2016-04-12
Software RenderMode does not support high resolution

breathedevelopment:
If you set Starling.supportHighResolutions to true on a high resolution screen, such as a MacBook Retina, while in Software mode you get an exception:

`exception, information=Error: Error #3780: Requested width of backbuffer is not in allowed range 32 to 2048.`

This fix also applies to Starling 1.8.

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -948,6 +948,8 @@ package starling.core
         public function get supportHighResolutions():Boolean { return _supportHighResolutions; }
         public function set supportHighResolutions(value:Boolean):void 
         {
+            if (context && context.driverInfo.indexOf("Software") != -1) return;
+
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Hm ... I'm not sure if we should prevent HiDPI mode in Software mode completely. E.g. a small Flash game in the browser might be smaller than 1024 points, and that would work, right? On the other hand, it would probably be super slow ... 

What about this: in "configureBackBuffer", I check if the maximum resolution is exceeded, and in that case I will fall back to the low DPI mode. Would that work for you?

breathedevelopment:
That sounds like a good solution. Thanks Daniel!

PrimaryFeather:
I have now implemented it like that! Please try it out, it did the trick in my tests. :smile:

PR #852, opened 2016-04-28
add Strength property to the filter

aberestovskyi:
Add strength property to the filters like in the native flash filter.

With this property you can make stroke for the text
flash_example

starling_example

var textField:TextField = new TextField(400,200, "Hello World");
textField.format.setTo("DR Agu", 60, 0xFF0000);
textField.x = 50;
textField.y = 50;
textField.filter = new GlowFilter(0x000000, 1, 0.5, 2.5, 1);
addChild(textField);

### DIFF
--- a/starling/src/starling/filters/BlurFilter.as
+++ b/starling/src/starling/filters/BlurFilter.as
@@ -20,6 +20,7 @@ package starling.filters
     {
         private var _blurX:Number;
         private var _blurY:Number;
+        private var _strength:Number = 1;
 
         /** Create a new BlurFilter. For each blur direction, the number of required passes is
          *  <code>Math.ceil(blur)</code>.
@@ -31,10 +32,11 @@ package starling.filters
          *      <li>etc.</li>
          *  </ul>
          */
-        public function BlurFilter(blurX:Number=1.0, blurY:Number=1.0, resolution:Number=1.0)
+        public function BlurFilter(blurX:Number=1.0, blurY:Number=1.0, strength:Number = 1.0, resolution:Number=1.0)
         {
             _blurX = blurX;
             _blurY = blurY;
+            _strength = strength;
             this.resolution = resolution;
         }
 
@@ -51,6 +53,7 @@ package starling.filters
                 return super.process(painter, helper, input0);
             }
 
+            effect.saturation = _strength;
             var blurX:Number = Math.abs(_blurX);
             var blurY:Number = Math.abs(_blurY);
             var outTexture:Texture = input0;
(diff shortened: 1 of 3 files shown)


PrimaryFeather:
Thanks a lot for the pull request, that looks super useful! I'll look into it as soon as time permits.

PrimaryFeather:
I've just made a simpler change that should do the trick, just as well!
If you pull the latest Starling version, you can pass alpha values > 1 to the constructor (or the `alpha` property). Any value above one will raise the "strength".

My recommendation for a nice outline:

```as3
new GlowFilter(0x0, 15, 0.5, 1.0);
```

(Use `resolution = 1` for best results.) Raise the alpha value even higher to make it sharper.

Does that work for you?

aberestovskyi:
Alpha trick doesn't work for me

PrimaryFeather:
Did you pull the very latest version? The change should be equivalent to what you did, actually.
https://github.com/Gamua/Starling-Framework/commit/119b853343f6b0057a5108ccec22365cafe626bc

aberestovskyi:
oh sorry, didn't pull. Yes, with your change it working. But big alpha doesn't good solution as for me. It will be unclear for people.

PrimaryFeather:
Yes, it definitely needs to be documented, that's for sure.

And yes, it might also be useful to add a specific property, like you suggested. But for now, at least we've got a working solution — I'll think about adding that. 👍

aberestovskyi:
Alpha trick doesn't work with not black colors http://c2n.me/3xRPrjr

//alpha trick
new GlowFilter(0x175591, 20, 1, 1);

//new strength property
new GlowFilter(0x175591, 1, 0.5, 14, 1);

PrimaryFeather:
Indeed, that's a good point!

sashkooo:
I think the strength property must be the last, because it will not break existing code which have a resolution property is already set.

aberestovskyi:
I used to native Flash filters for the text befor Starling 2.0. And I set this values befor resolution, because native flash filters has such order. It's easier to switch native flash filters to Starling flash filters. And as for me, filter looks much better with resolution 1.

sashkooo:
for replace native filters yes, but anyway native filters must be reconfigured. Instead, projects that already uses Starling filters will not need any changes. For example, I have in my project already 70 places where Starling filters are used, and 80 places where nativeFilters are used.

PrimaryFeather:
I now fixed the color problem you reported a while ago; with this change, the `alpha` value should act as a valid replacement for `strength`.

Actually, I would have preferred to add a real `strength` parameter — but as you noticed, I would have needed to add this parameter after the `resolution`, i.e. the order would not have been the same as in the classic Flash class. (Changing the order is not an option, because it would break other people's code.)

So, I decided (for now) to keep the order and naming just the same, and just document the additional usage of the `alpha` parameter. In a future version, I might split it up instead — but for now, you've got a working solution.

PR #916, opened 2016-10-18
Add color component setter functions

oddwhocanfly:
It makes very easy to change single component of existing color.


PrimaryFeather:
Thanks a lot for the pull request! You're right, that makes perfect sense — should have been in there all along! :smile:

PR #917, opened 2016-10-19
Letterspacing property in TextFormat

pbalmasov:
added letterSpacing to TextFormat


PrimaryFeather:
You're right, that was still missing. 👍

PR #925, opened 2016-11-17
On Windows, setRequiresRedraw() is required when resizing view port or Starling stage after resizing NativeWindow and skipUnchangedFrames is true

joshtynjala:
I was able to confirm the following issue reported by a couple of users in the forums. It worked correctly on my Mac, but when I tried on Windows 10, increasing the dimensions of the NativeWindow resulted in white regions on the edges. I'm not sure why it works on one platform, and not the other.

The following post in the thread includes a screenshot and some MXML code for the Feathers SDK that will reproduce the issue (I'm sure it's easy to reproduce purely with Starling and another SDK, though):

http://forum.starling-framework.org/topic/feathers-background-not-resizing/page/2#post-94362

When I added calls to setRequiresRedraw() either to Starling's viewPort property or to Stage's stageWidth and stageHeight properties, I could no longer reproduce the issue. For safety, I think the call to setRequiresRedraw() should be included in both places, because there may be a rare situation where a developer might change the view port or the stage dimensions, but not both.

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -797,7 +797,11 @@ package starling.core
         
         /** The viewport into which Starling contents will be rendered. */
         public function get viewPort():Rectangle { return _viewPort; }
-        public function set viewPort(value:Rectangle):void { _viewPort = value.clone(); }
+        public function set viewPort(value:Rectangle):void
+        {
+            _viewPort = value.clone();
+            _stage.setRequiresRedraw();
+        }
         
         /** The ratio between viewPort width and stage width. Useful for choosing a different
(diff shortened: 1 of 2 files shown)


PrimaryFeather:
I just noticed that I could avoid the `clone` call in the `set viewPort` property, as well as move `setRequiresRedraw` to a different place to make sure it also happens when someone does not assign a new viewPort, but instead changes it directly (e.g. `starling.viewPort.width = ...`). I committed that change, too.

PrimaryFeather:
In any case: thanks a lot for looking into it and for creating the pull request! :smile:

joshtynjala:
Thanks, Daniel!

PR #926, opened 2016-11-18
Get application version number setup in the *-app.xml

alamboley:


PrimaryFeather:
Good idea, that's definitely helpful!
One suggestion: could you rename the property to just `applicationVersion`? For consistency with the other `version` property.
Thx!

PrimaryFeather:
One problem I just realized is that such a version information is only available in AIR. All other properties in that class currently work for both AIR and Flash. But there's no sensible fallback value for Flash in this case. 😕

alamboley:
Hi Daniel,
I just renamed the property. Yeah, it's true. Unfortunately there is no way to have the same info with FP.

PrimaryFeather:
Thanks for the update, Aymeric! 

> Yeah, it's true. Unfortunately there is no way to have the same info with FP.

Mhm, there's no way around that. However, it makes me wonder if this is the right place for this property. Plus, people might want other information from the application descriptor, too — and it would be weird if there were some day lots of properties in the SystemUtil that just wrap lookups into the XML.

It's actually a little weird that this property is not part of the `flash.desktop.nativeApplication` class (like `applicationID`)!

So I'm not entirely sure about it yet.

PR #932, opened 2016-12-06
Small MeshBatch Fix

racarone:
I've created an Effect that depends on the state of its parent Style for program creation. This fix just ensures that the style is copied _before_ a new style is set on a batch.

### DIFF
--- a/starling/src/starling/display/MeshBatch.as
+++ b/starling/src/starling/display/MeshBatch.as
@@ -187,9 +187,15 @@ package starling.display
             var meshStyleType:Class = meshStyle.type;
 
             if (_style.type != meshStyleType)
-                setStyle(new meshStyleType() as MeshStyle, false);
-
-            _style.copyFrom(meshStyle);
+            {
+                var newStyle:MeshStyle = new meshStyleType() as MeshStyle;
+                newStyle.copyFrom(meshStyle);
+                setStyle(newStyle, false);
+            }
+            else
+            {
+                _style.copyFrom(meshStyle);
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Thanks a lot for the pull request, Robert! Happy to see that you're working with styles! :smile:

I can totally merge that, to avoid any unnecessary initializations — but just to be sure: you're aware of the `updateEffect` method, right? That's called before each rendering, and should be used to configure the effect according to it's style's settings. So this should — in theory — have solved the problem, as well. (If I'm understanding the problem correctly?)

racarone:
Yup, getting deep into the framework! 😁

Ya I am, but it in a sense feels wrong setting an `Effect` initialization parameter repeatedly when it's only required once during creation...

Basically I'm writing a shader heavy game and I got tried or creating an `Effect`, `Style` and then duplicating their properties for every shader combination, so I came up with a slightly more dynamic approach. It uses a single `Effect` class that has a `ConstantData` class for the fragment and vertex shaders. It knows the layout of a shader, sort of like `VertexData`, and sets the attributes, textures, and constants accordingly. The style has methods for setting named constants and then just updates the effect in `updateEffect`. I pass the shader and layout to the Effect from the style on creation, instead of every frame. It's not completely fleshed out, but it's sped up my shader development quite a bit. =)

PrimaryFeather:
That sounds like a great idea — keep it up! :smile:

racarone:
Thanks Daniel! =)

PR #935, opened 2016-12-18
initialized helper matrix to reduce allocations

bwhiting:
Noticed a number of Matrix3D allocations in filters and thought this should have been initialized.


PrimaryFeather:
Great find, thanks a lot! Indeed, that was the way I actually intended using this matrix. 😬 
Thanks!

PR #955, opened 2017-04-05
Bug: double dispose call if text is empty

5k0rp:
fillMeshBatch does not replace texture if text is empty


PrimaryFeather:
Thanks a lot for the bugfix! I merged it right away. :smile:

PR #958, opened 2017-04-26
Add static functions to Image for binding a scale9Grid to a Texture

kevinfoley:
The scale9Grid will automatically be applied whenever the Texture is displayed in an Image, saving us the boilerplate of passing around the scale9Grid with the Texture.

Usage:

    var texture:Texture = assetManager.getTexture("scale9-texture");
    var scale9Grid:Rectangle = new Rectangle(4, 4, 56, 56);
    //call this once after your textures have loaded
    Image.bindScale9GridToTexture(texture, scale9Grid);

    var image:Image = new Image(texture);
    var button:Button = new Button(texture, "Press me");

### DIFF
--- a/starling/src/starling/display/Image.as
+++ b/starling/src/starling/display/Image.as
@@ -1,16 +1,17 @@
 // =================================================================================================
 //
-//	Starling Framework
-//	Copyright Gamua GmbH. All Rights Reserved.
+//    Starling Framework
+//    Copyright Gamua GmbH. All Rights Reserved.
 //
-//	This program is free software. You can redistribute and/or modify it
-//	in accordance with the terms of the accompanying license agreement.
+//    This program is free software. You can redistribute and/or modify it
+//    in accordance with the terms of the accompanying license agreement.
 //
 // =================================================================================================
 
 package starling.display
 {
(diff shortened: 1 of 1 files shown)


PrimaryFeather:

I actually really like this idea!! Automatically applying the scale9 settings is something that should really be automated. And your approach has the advantage that it doesn't pollute the _Texture_ class.

I'm actually considering making this even more versatile. E.g. sometimes a texture also indicates a certain pivot point. Perhaps we can use your "binding" approach for that, as well — or change it to one technique (a callback, maybe?) that solves both (and more) issues.

I'm currently busy with finally getting the "Starling Handbook" out of the door, but that's really in the final stage. I'll review this when it's out!

Thanks again for this great suggestion, Kevin!!

PrimaryFeather:
That's now part of the master branch!

I actually added a generic function called `automateSetupForTexture` that can be used to execute arbitrary code whenever a specific texture is assigned. Then, I added two convenience methods that bind a `scale9Grid` or a pivot point to a given texture.

Thanks again for the idea and for the pull request! I hope you're fine with this implementation, as well. :smile:

PR #960, opened 2017-04-30
Employing the `numSides` parameter for elliptical gen methods

EhsanMarufi:
The useful parameter of `numSides` in the constructor of the internal class of `Ellipse` is implemented and available; why not allow the users to benefit the advantages of the parameter?!
As the class provides a resolution-independent concept, one may need to have more control on the sides' count, e.g. where the shape may be scaled in a way the imperfection of the to-be elliptical curve is undesirable, and the issue could've been solved easily, if there were more sides on the shape!

### DIFF
--- a/starling/src/starling/geom/Polygon.as
+++ b/starling/src/starling/geom/Polygon.as
@@ -283,15 +283,15 @@ package starling.geom
         // factory methods
 
         /** Creates an ellipse with optimized implementations of triangulation, hitTest, etc. */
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
You're perfectly right, it makes sense to make that parameter generally available. To be honest, I don't know why I didn't do it in the first place ... 

Thanks a lot for taking the time to set up the pull request!!

PR #963, opened 2017-05-03
SystemUtil: added convenience methods

kevinfoley:
Added `isAndroid`, `isIOS`, `isMac`, `isWindows` as shortcuts for manually checking the `platform` string

### DIFF
--- a/starling/src/starling/utils/SystemUtil.as
+++ b/starling/src/starling/utils/SystemUtil.as
@@ -115,6 +115,12 @@ package starling.utils
             initialize();
             return sAIR;
         }
+		
+		/** Indicates if the code is executed on an Android device, based on the <code>platform</code> string.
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
You're right, that makes sense — I'll add that!

Could you do me a favour and change the code style so that it fits the rest of Starling's source? 

* Move curly braces into their own line
* Use 4 spaces instead of tabs for indentation

I know I'm a nitpicker, but I'd like everything to be consistent. ;-)
Thanks in advance!!!

kevinfoley:
Whoops, guess I was going too fast when I put that together! I've fixed the formatting to match your standard style.

PrimaryFeather:
Thanks, Kevin! That's much better. :smile:
Merged!

PR #976, opened 2017-06-08
Introducing the Star polygon

EhsanMarufi:
Using the new `Star` class, you can create custom n-pointed star shapes with _really_ optimized triangulation and `hitTest()` functionalities.
The _Starling_ framework is no longer imaginable without the polygonal _star_ generation routine _!!_ :joy: (joking)

### DIFF
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -70,6 +70,12 @@ package starling.display
             appendPolygon(Polygon.createEllipse(x + radiusX, y + radiusY, radiusX, radiusY));
         }
 
+        /** Draws an n-pointed star. */
+        public function drawStar(x:Number, y:Number, radius1:Number, radius2:Number, numPoints:uint, angleOffset:Number = 0):void
+        {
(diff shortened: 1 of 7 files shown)


EhsanMarufi:
Oops!!! The Git was supposed to compare the commits and create a new pull request based on the new commits, but the previous commits are also _here!!_ What am I suppose to do?!!

PrimaryFeather:
Don't worry about it, Ehsan, I know it's sometimes confusing how GitHub handles these things. :wink:

In any case, in this particular situation, I'd rather have this feature as an [extension](http://wiki.starling-framework.org/extensions/start), e.g. simply one class (subclassing _Polygon_) that can be used by importing it into one's project. I'm  still not 100% sure how I'm going to move on with the vector capabilities, so I want to keep this feature set really small for now.

EhsanMarufi:
> it's sometimes confusing how GitHub handles these things..

I know, it happens every once in a while :blush:

> I'm still not 100% sure how I'm going to move on with the vector capabilities

Hmm.. It's OK, I'm sure you're gonna find a very convenient approach to these capabilities, you will. But until then, you may utilize the optimizations applied in the code for the hit test routine of the `Star`.
Thank you Daniel :)

PR #987, opened 2017-07-14
Check context in Painter before disposing

Liklainy:
Have a problem with Starling's initialisation on iOS.
It fails with error 3702 on context initialisation if application was in background.
To recreate this error you need to move application into background when Starling is waiting for context creation.
It can be done many different ways, but the simplest is by double pressing home button.
After that a little waiting is needed, and select the application again.
It's like in one of the forum's posts:
https://forum.starling-framework.org/topic/recover-from-context3d-not-available-when-starting-ios-app-in-the-background
Was this already solved before?

For now I do workaround, by catching ERROR and recreating Starling instance.
And I tried to dispose old instance to ensure that it will not catch stage events.
But failed on Painter's dispose method, because there was no context.

### DIFF
--- a/starling/src/starling/rendering/Painter.as
+++ b/starling/src/starling/rendering/Painter.as
@@ -166,7 +166,8 @@ package starling.rendering
 
             if (!_shareContext)
             {
-                _context.dispose(false);
+                if (_context)
+                    _context.dispose(false);
                 sSharedData = new Dictionary();
             }
         }
--- a/starling/src/starling/utils/RenderUtil.as
+++ b/starling/src/starling/utils/RenderUtil.as
@@ -232,7 +232,11 @@ package starling.utils
             {
                 currentProfile = profiles.shift();
 
-                try { execute(stage3D.requestContext3D, renderMode, currentProfile); }
+                try
+                {
+                    SystemUtil.executeWhenApplicationIsActive(
+                        stage3D.requestContext3D, renderMode, currentProfile);
+                }
                 catch (error:Error)
                 {
                     if (profiles.length != 0) setTimeout(requestNextProfile, 1);


PrimaryFeather:
Thanks a lot for the pull request and the report — and sorry that it took so long to get back to you!

I haven't yet managed to reproduce the problem, though. Could you elaborate on what exactly I have to do to get to this error? I must be doing something wrong.

PrimaryFeather:
Ah, now I could reproduce it. It only happens when background execution is enabled, of course. :bowtie: 

I'll have a look at it!

PrimaryFeather:
I just committed a change that might catch the problem at its root: when trying to create the context, Starling now checks if the application is active, and delays context creation if it's not.

Caveat: this only works when the Starling *knows* that the application is in the background. Unfortunately, there is no AIR API that tells us this information directly (at least none that I know of), so Starling can only listen to ACTIVATE and DEACTIVATE events, and remember the status.

Thus, to be on the safe side, I also recommend to add the following call to the *very beginning* of your app:

```as3
SystemUtil.initialize();
```

This will make sure Starling immediately starts to listen for (DE)ACTIVATE events.

Please try it out and let me know if that works for you!

Liklainy:
Thank you! Seams to be working for me.

PrimaryFeather:
Fantastic, I'm happy to hear that! Thanks for trying it out. :smile:

barbudumitruadrian:
Hi Daniel,

Unfortunately, this fix creates another issue:
I'm running on a Desktop app, make a delay to reload Starling in 1 second, then I deactivate the Starling window. This will lead that the Context3D will not be created right away, but only after I click again the window app (thus activating it). My internal fix was to change RenderUtil to execute directly stage3D.requestContext3D if SystemUtil.isDesktop.

Thanks.

PrimaryFeather:
You are perfectly right, this logic needs to be limited to mobile platforms! I just made sure that's the case, just as you suggested.

Thanks a lot for the heads-up, @barbudumitruadrian!

barbudumitruadrian:
Uuuu, lightning fast response and fix.
Thanks, Daniel.

PR #1028, opened 2018-05-08
Fast Memory VertexData implementation

aram-ahak:
We have implemented Domain Memory based fast memory utilization for vertex data.
Results are VERY promising, and we would love to have all of the community pushing these changes forward.
In our project we observe 20-30% rendering speed increase. The smoothness of the app also is observed by eye of user.

@PrimaryFeather we hope you will have a look very soon and will benchmark these changes against current master branch.


JohnBlackburne:
What issues have you encountered in implementing this? I ask as I have looked into this a couple of times myself, and could see a couple of problems. First how do you decide how much Domain Memory to allocate? For any particular app it might be possible to decide on an allocation, but it would be hard to come up with a general approach. Second how do you deal with memory fragmentation? Again, this is something that might not be an issue for an individual app, but which is needed for a general solution.

gevorg-gsar:
@JohnBlackburne we allocate more memory if it gets filled. We have defragmentation logic integrated with allocation logic, this allows to allocate nearest vacant memory chunk.
The solution works fine for our quite large app! and we have really good performance boost because of this.
Now we need everyone to jump on this train, try & test, and work collaboratively on making this better.

JohnBlackburne:
Looking more closely at the code it looks like it has a fixed size of 1MB, with no scope for increasing that if you run out of space. And it looks like it does nothing to counter fragmentation due to chunks that are too small to use, arising as a chunk is freed then re-used for something a little smaller, accumulating over time. These may be constraints you can live with but they are far less general than what is in Starling right now.

gevorg-gsar:
@JohnBlackburne yes, let's work on making this more general. 
50% rendering performance increase is a great motivator. We can agree on one thing, that we can't expect such performance boost from Adobe.

gevorg-gsar:
@PrimaryFeather thanks for your reply! 

We understands the concerns you have about moving this changes to the main branch. 
Having it as a fork is fine. Our goal is to put community behind this approach and make it even better.

So far about memory I must say that we are working on making the allocated memory size basically unlimited  (as far as it is allocatable from OS). We are writing a lazy defragmentation logic for that. 
Currently we allow to up to 10MB memory allocation, with this change it will be as much as necessary.

No problem also about changing the code to your standards. 
We will work on that too. 
Meanwhile we will be happy to see more people behind this fork and its future improvements. I suggest to make a new threads to discuss this in the forum as well.

gevorg-gsar:
Hi! 
We  just added changes with following fixes:

- 2 memory leaks fixed
- memory defragmentation logic added allowing to free memory when allocated memory passes threshold of 10MB
 
Now we allow basically any available size memory allocation, without any memory waste by having defragmentation logic in place.

PrimaryFeather:
Awesome! Thanks for your continuing work on this, and also for creating the [thread](https://forum.starling-framework.org/topic/developer-to-benchmark-our-starling-fastbytearray-vertexdata-implementation) in the forum!

PR #1031, opened 2018-06-11
Signature of error callback function

aram-ahak:
Hi guys.
In our app we need to know information about the failed asset.
Is it possible to add the failed asset in error callback function?


PrimaryFeather:
Excellent idea! You're right, that's important to actually act on the error. Thanks!

PR #1032, opened 2018-06-12
Starling: calls setRequiresRedraw() on Event.ACTIVATE because it is sometimes required with AIR on Windows

joshtynjala:
If a desktop AIR app is minimized, it may render with a blank white screen when the window is restored later. Moving the mouse around may cause Starling to start rendering again if the interaction triggers anything that might require a redraw.

There are a few requirements to reproduce this issue:

1) **Windows** only. The issue does not reproduce on macOS.

2) Must be using **Context3DProfile.BASELINE_CONSTRAINED**.

3) **skipUnchangedFrames** must be true.

4) Must **NOT** stop Starling on Event.DEACTIVATE and restart it on Event.ACTIVATE.

To fix this issue, I added a listener for Event.ACTIVATE inside the Starling class. In the listener, I call setRequiresRedraw() after a 100 ms delay. Calling it immediately did not work for me. This is the same workaround that already exists in start() (the comment says it may be required on Android), but it works even if you want to allow a desktop app to render while in the background.

I did not remove the existing workaround in start(), but I think it may be redundant now.

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -300,6 +300,7 @@ package starling.core
             stage.addEventListener(KeyboardEvent.KEY_UP, onKey, false, 0, true);
             stage.addEventListener(Event.RESIZE, onResize, false, 0, true);
             stage.addEventListener(Event.MOUSE_LEAVE, onMouseLeave, false, 0, true);
+            stage.addEventListener(Event.ACTIVATE, onActivate, false, 0, true);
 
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Thanks a lot for the pull request, Josh! Merged right away. I'll also remove the other workaround, because, as you say, it's now redundant.

joshtynjala:
Thanks, Daniel!

PR #1043, opened 2018-09-10
Fix RenderTexture dispose() texture

Doppp:
_activeTexture was being disposed twice once in RenderTexture and once in SubTexture. Only dispose it in RenderTexture if SubTexture isn't going to dispose it.

### DIFF
--- a/starling/src/starling/textures/RenderTexture.as
+++ b/starling/src/starling/textures/RenderTexture.as
@@ -143,7 +143,10 @@ package starling.textures
         /** @inheritDoc */
         public override function dispose():void
         {
-            _activeTexture.dispose();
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Thanks a lot for the pull request! Indeed, that `dispose` should not be called twice, as it is now — good catch!

I'd even go so far as to remove that `dispose` call inside _RenderTexture_ completely. The active texture is always created in the constructor via `Texture.empty()` — and that returns either a concrete texture, or a sub-texture that owns its parent. So I'll remove that call completely.

Thanks again!

PrimaryFeather:
... and I realized it needs to be done in a slightly different way (see above commit). :wink:

Doppp:
Thanks very much, @PrimaryFeather!

PR #1069, opened 2020-02-14
updated to copy BitmapData with copyPixels();

Dimensionscape:
Update to fix issue with BitmapData distortion on DisplayObjects larger than the stage.

### DIFF
--- a/starling/src/starling/display/DisplayObject.as
+++ b/starling/src/starling/display/DisplayObject.as
@@ -1,4 +1,4 @@
-// =================================================================================================
+﻿// =================================================================================================
 //
 //	Starling Framework
 //	Copyright Gamua GmbH. All Rights Reserved.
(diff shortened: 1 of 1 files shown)


PrimaryFeather:

I hadn't heard about this problem yet – so you say when you use "drawToBitmapData" on Windows, with an image size bigger than the backbuffer, some parts are upside down?

Dimensionscape:
Yes, precisely. This occurs on Windows at least. I haven't checked on other platforms. Depending on the size of the stage, sometimes it works, sometimes it renders some of the chunks upside down. I'm not sure why the original code wasn't working, but assembling entire bitmapdata out via copyPixels() fixed it.

PrimaryFeather:
Thanks a lot for the clarification! I'll have a look at it in the coming days. 👍

PR #1076, opened 2020-07-10
Update AtfTextureFactory.as to execute custom TextureOptions.onReady

esidegallery:
A reference to the original TextureOptions.onReady is now stored and executed by the overwriting onReady closure.


PrimaryFeather:
You're right, that makes a lot of sense! Thanks a lot for taking the time to create the pull request. Happily merged! 😁

PR #1077, opened 2020-07-10
Update BitmapTextureFactory.as to execute custom TextureOptions.onReady

esidegallery:
A reference to the original TextureOptions.onReady is now stored and executed in complete(), which overwrites onReady.


PrimaryFeather:
Thanks again! A really useful change.

PR #1086, opened 2021-06-08
added small optimization to Polygon 'contains' method

Adolio:
Hi @PrimaryFeather!

Adobe Scout reported me that quite a lot of time was spent in the `numVertices` accessor while using the `contains` method.

What do you think about this small optimization? It seems to be used in few other similar methods too...

Best,
Aurélien


PrimaryFeather:
That's a great find, Aurélien! This is definitely worth the change – it should have been like that all along. 
Thanks a lot for the pull request! Merged right away. 😄

PR #1090, opened 2022-01-06
Asset Manager: Added detection of extension & mime type for embed assets

Adolio:
Hi Daniel,

Here is a proposal to enable the asset factories for embed assets. No API should be broken, I managed to add extra optional parameters to enforce URL and/or mime type.

👉 In order to enable this feature, an extra compiler option is required to keep the "Embed" metadata to access it from the `AssetManager`.

Thanks in advance for your review.

Best,
Aurélien

### DIFF
--- a/starling/src/starling/assets/AssetManager.as
+++ b/starling/src/starling/assets/AssetManager.as
@@ -71,15 +71,15 @@ package starling.assets
      *  <listing>
      *  var appDir:File = File.applicationDirectory;
      *  var assets:AssetManager = new AssetManager();
-     *  
+     *
      *  assets.textureOptions.format = Context3DTextureFormat.BGRA;
      *  assets.enqueue(appDir.resolvePath("textures/32bit"));
-     *  
+     *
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
Hey Aurélien,
thanks a lot for the pull request! This is looking really useful!

I'll review it in detail soon, and then merge and/or add some comments. 😄

Adolio:
Thanks Daniel, I'm looking forward to have your feedback :)

PrimaryFeather:
Great work, Aurélien! That's a great addition and simplifies usage of embedded assets greatly. Thanks again!!

In a follow-up commit, I moved the duplicated child processing logic into a local method to remove code duplication; but that's just me being overly pedantic. 😜 (See https://github.com/Gamua/Starling-Framework/commit/8dab458414c20b85015e1f126ebd8de3ec66c3ea.)

Adolio:
Thanks Daniel!

Yep, your extra commit does completely make sense to me as well 😄 Well done 👍

PR #1092, opened 2022-01-06
Starling: Added a default texture smoothing option to configure it globally

Adolio:
Hey Daniel,

I have added a global parameter in the `Starling` class to override the default texture smoothing value which is used in `MeshStyle`, `FragmentFilter` & `FilterEffect`.

This allows for example to setup "trilinear" texture filtering everywhere by doing:

```as3
var myFavouriteBird:Starling = new Starling(...);
myFavouriteBird.textureSmoothing = TextureSmoothing.TRILINEAR;
```
☝️ Note: this doesn't update any pre-existing meshes & filters.

Best,
Aurélien

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -44,6 +44,7 @@ package starling.core
     import starling.events.TouchPhase;
     import starling.events.TouchProcessor;
     import starling.rendering.Painter;
+    import starling.textures.TextureSmoothing;
     import starling.utils.Align;
     import starling.utils.Color;
     import starling.utils.MatrixUtil;
@@ -53,10 +54,10 @@ package starling.core
 
     /** Dispatched when a new render context is created. The 'data' property references the context. */
     [Event(name="context3DCreate", type="starling.events.Event")]
-    
+
     /** Dispatched when the root class has been created. The 'data' property references that object. */
     [Event(name="rootCreated", type="starling.events.Event")]
-    
+
     /** Dispatched when a fatal error is encountered. The 'data' property contains an error string. */
     [Event(name="fatalError", type="starling.events.Event")]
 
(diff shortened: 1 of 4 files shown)


PrimaryFeather:
I can totally see that this might be useful – thanks for the pull request, Aurélien!

I'd rather name this new property `defaultTextureSmoothing`, though. That way, it's more apparent that this is just a default that can be overridden anytime. I admit it's a long word to type, but I still think it's worth it. :wink:

What do you say?

Adolio:
I don't have any concern about renaming this variable, I'm usually the one which is complaining about variable & class names 😅 In this particular case I just wanted to mimic what was done for the `antiAliasing` but understand that maybe it's slightly different since it's the "value" used as default.

Adolio:
I will do the changes ASAP.

PrimaryFeather:
Hm, that similarity with `antiAliasing` is a good point. Give me a minute, I'll have another look!

PrimaryFeather:
Okay, the difference is that `Starling.antiAliasing` really only influences the back buffer, which is handled by the Starling class. What you set here is actually used by Starling. 

However, the `textureSmoothing` property is used by _other_ classes and can be overridden at multiple places. So in that light, I think it makes sense to add that `default`  prefix.

Thanks for mentioning it, though! I hate it when the API is inconsistent. 😉

Adolio:
Perfect, this makes sense too me as well - by the way, I initially wanted to name it exactly that way 😉

PrimaryFeather:
> Perfect, this makes sense too me as well - by the way, I initially wanted to name it exactly that way 

Haha, brilliant! Then go for it! 😄

Adolio:
Here we go 🙂!

Adolio:
Sorry I forgot to stage two files - the last commit is fine now.

PrimaryFeather:
Perfect – thanks again! It's now part of the master branch. 😄

PR #1093, opened 2022-01-14
Optimize removeChild

HaimZik:
Prevent RemoveChild needlessly doing indexOf.
This was causing O(n log n) behavior in the case of removing all children.
Fix short black screen at the end of Starling demo benchmark for large amount of Images.


PrimaryFeather:
Thanks a lot, a very useful optimization! 👍

PR #1099, opened 2022-07-04
Juggler: Reduced memory allocations by removing the Dictionary implementation for object ids

Adolio:
This should prevent many String allocations

**Changes**
- The `_objectIDs` is now a `Vector.<uint>` containing the objects' ids
- The cells cleaning logic as been factorized by using the new `removeByIndex` method
- `object in _objectIDs` have been replaced by `contains(object)`
- The cells re-oganization (in the `advanceTime` method) logic has been updated accordingly

This PR needs further testing. Changes are working on my side but I'm not sure I have covered all cases.

### DIFF
--- a/starling/src/starling/animation/Juggler.as
+++ b/starling/src/starling/animation/Juggler.as
@@ -10,8 +10,6 @@
 
 package starling.animation
 {
-    import flash.utils.Dictionary;
-
     import starling.core.starling_internal;
     import starling.events.Event;
     import starling.events.EventDispatcher;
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
This is really an exemplary example for a pull request, Aurélien! Thanks a lot!

I really like how you got rid of the ugly code duplication I had in there whenever an object was removed. I'm actually a little embarrassed that it wasn't implemented like that originally! 🫢

Thankfully, the Juggler class is extensively unit tested – there are a few traps in there which can easily be overlooked, but I think all of them are covered by the unit tests. And those are running just fine after merging your changes.

So, everything is looking great! Thanks a lot for your careful implementation, Aurélien! 😄👏

Adolio:
Thanks Daniel for your careful review 😁

I'm very happy that you liked it, this means a lot to me coming from you! ☺️

PR #1100, opened 2022-07-28
Json Factory: Added support for root json array

Adolio:
Hi Daniel,

I just noticed that the `Json Factory` doesn't support root JSON array. 

According to the format definition root arrays are valid. So after this changes embed files starting with `[` will be considered as JSON and will be parsed accordingly.

I tested it and everything works as expected.

See you,
Aurélien


PrimaryFeather:
You're totally right, Aurélien, nothing speaks against supporting arrays here, as well. Thanks for creating the pull request!! 😄👏

Adolio:
Thanks Daniel for the merge! ☺️

PR #1102, opened 2022-09-11
Update AssetManager.as

BladePoint:
Allow subclasses access to NAME_REGEX matching.


PrimaryFeather:
Agreed, that's a useful addition. Thanks!

PR #1105, opened 2023-01-03
Starling: Fix ambiguous reference compiler warning

Adolio:
This PR fixes the following compiler warning:

ambiguous reference warning

@PrimaryFeather Here is my renaming proposal, please let me know if you think that another name would be better.


Adolio:
I don't know exactly because the integration of this feature has been delayed a bit on my side. It might also come from AIR 50.x 🤷‍♂️

Thanks for the merge Daniel 🙂

PR #1106, opened 2023-03-21
fix for BlendMode.isRegistered function error when calling it early

MatseFR:
if you call this when your main Starling class is created or added to stage, it results in an error because sBlendModes is null. Other BlendMode methods check for this and call registerDefaults() if needed so I did the same


PrimaryFeather:
Ah, that's a good find — thanks for the quick fix, Mathieu! 🙏

PR #1109, opened 2024-07-13
Starling: Small rendering events improvements

Adolio:
Hi @PrimaryFeather !

Here is a small PR to propose an improvement regarding the rendering events.

### Changes

1) The first commit introduces a new event `RENDER_COMPLETE` to indicate when the rendering has been completed.
2) The second commit fixes the stats display to take into account all draw calls that might happen during the reception of the RENDER event.

### Context

I'm using those changes to implement a post processing stack at different stages of rendering (screen, camera, etc.). The first commit let me hide display objects (mark them as not visible) and make them back visible after rendering to let them be taken into account for the touch processing.

The second commit allows me to track all draw calls done during the deferred rendering through render textures.

I'm obviously open to discuss any side effect you might see from those changes.

Best,
Aurélien

### DIFF
--- a/starling/src/starling/core/Starling.as
+++ b/starling/src/starling/core/Starling.as
@@ -429,7 +429,8 @@ package starling.core
          *
          *  <p>This method also dispatches an <code>Event.RENDER</code>-event on the Starling
          *  instance. That's the last opportunity to make changes before the display list is
-         *  rendered.</p> */
+         *  rendered. When the rendering is completed, this method dispatches an
+         *  <code>Event.RENDER_COMPLETE</code>-event on the Starling instance too.</p> */
         public function render():void
         {
             if (!contextValid)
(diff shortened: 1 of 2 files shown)


PrimaryFeather:
Thanks for the PR, Aurélien! That sounds really useful – I'll review it within the next few days. 👍

PrimaryFeather:
Hey @Adolio,
the new event sounds useful, I can totally see your point! And moving `_painter.nextFrame()` to the beginning of the method  (before dispatching the event) is probably where it should have been all along. It might theoretically have side effects if someone relied on the old behavior, but it's unlikely. 

So I'll gladly merge the change! Thanks again! I'm really happy to see Starling is still useful to you! 😎

Adolio:
Thanks @PrimaryFeather for merging 🤩

Absolutely, I'm still thoroughly enjoying exploring this ecosystem, and I'm continually impressed by the work you've done here, even after all these years! The Starling Framework is such a great place to dive into 🤗

PrimaryFeather:
Thanks for the nice words, Aurélien! 
Working on Starling was always a great pleasure, and I'm still feeling privileged I could do it for such a long time. 😊

PR #1110, opened 2024-09-13
Display Object Container: Fixed bounds computation when a container has children with no size

Adolio:
Hey @PrimaryFeather,

I just discovered a surprising issue related to the computation of bounds of display container composed with "non sized" display objects. Those "non sized" display objects contribute to the bounds of the parent but actually shouldn't.

This results with incorrect bounds.

---

Here is a simple code repoducing the issue:

```
var mainSprite:Sprite = new Sprite();
trace("Sprite width: " + mainSprite.bounds.width);

var emptySprite:Sprite = new Sprite();
mainSprite.addChild(emptySprite);
trace("Sprite width: " + mainSprite.bounds.width);

var quad:Quad = new Quad(10, 10);
quad.x = 100;
mainSprite.addChild(quad);
trace("Sprite width: " + mainSprite.bounds.width);
```
The expected width is `10` since only the quad is visible.

---

Output without the fix:
```
Sprite width: 0
Sprite width: 0
Sprite width: 110
```
👉 This is due to the fact that the empty sprite is located at 0 influencing the bounds from 0 to 100 (quad's position) + 10 (quad's size).

Output with the fix:
```
Sprite width: 0
Sprite width: 0
Sprite width: 10
```

### DIFF
--- a/starling/src/starling/display/DisplayObjectContainer.as
+++ b/starling/src/starling/display/DisplayObjectContainer.as
@@ -305,6 +305,7 @@ package starling.display
             }
             else
             {
+                var sizedChildrenCount:uint = 0;
                 var minX:Number = Number.MAX_VALUE, maxX:Number = -Number.MAX_VALUE;
                 var minY:Number = Number.MAX_VALUE, maxY:Number = -Number.MAX_VALUE;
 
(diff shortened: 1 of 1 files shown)


PrimaryFeather:
@Adolio 

Hey Aurélien, 
first of all: sorry for the late reply, I had a few busy days. 😉 

Hm, that's interesting! You're right, your fix is very likely the way to go (thus I merged it right away). I have some dark recollection of this having been reported long ago and me answering "it's as designed", but I can't for the world figure out a good reason why it should behave like this. 😆 So let's just add this fix and if problems should arise, we'll notice for sure.

It's definitely e.g. a good idea for sprites with a fragment filter, because this can potentially reduce the size of the filter texture. 

Thanks for digging again and again into the code! 😁🙏 

Cheers,
Daniel

Adolio:
@PrimaryFeather

Thank you so much for considering and merging the proposed changes! I hope they'll have a positive impact on existing projects 🤞. On my end, everything is still working as expected.

No worries about the slight delay – I wasn't expecting such a quick response, but it was greatly appreciated!

It's always a pleasure to dig into the project 🤗.

See you,
Aurélien

--- 

By the way, do you have any plans to release a minor version of Starling soon or later to include the small fixes and improvements since `2.7`? Do you think it would be worth it?

PR #1115, opened 2024-10-02
Update ConcretePotTexture.as

zach-xhx:
When uploading a 2048x2048 texture, all smaller mipmap levels were then uploading as 1024x1024, because that was the only BitmapData size that ever got created in the whole mipmap upload process. I changed it so now the deeper levels upload as BitmapData objects that are sized 512x512, 256x256, etc. Was it correct for me to change this, or am I making a mistake?

Also I am curious, why doesn't Starling support async uploads for mipmaps, is it due to hardware limitations or something else?

### DIFF
--- a/starling/src/starling/textures/ConcretePotTexture.as
+++ b/starling/src/starling/textures/ConcretePotTexture.as
@@ -89,6 +89,7 @@ package starling.textures
             {
                 var currentWidth:int  = data.width  >> 1;
                 var currentHeight:int = data.height >> 1;
+                var last_canvas:BitmapData = null;
                 var level:int = 1;
                 var matrix:Matrix = sMatrix;
                 matrix.setTo(0.5, 0.0, 0.0, 0.5, 0.0, 0.0);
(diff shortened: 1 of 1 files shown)


zach-xhx (commented):
Please disregard this update and review the next update I submitted afterwards, it includes a much more important change. Sorry, I am not too experienced with github, I hope I am doing this right.

PrimaryFeather:
Hey there, @zach-xhx!

Those changes are looking totally reasonable to me! I'm really questioning why I originally implemented it in any other way. 

> When uploading a 2048x2048 texture, all smaller mipmap levels were then uploading as 1024x1024, because that was the only BitmapData size that ever got created in the whole mipmap upload process. I changed it so now the deeper levels upload as BitmapData objects that are sized 512x512, 256x256, etc. Was it correct for me to change this, or am I making a mistake?

My guess is that I wanted to avoid having to create multiple BitmapData objects and thought that the `texture.upload` method would only look at the part that's required. After all, it does display the mipmaps correctly! So I'm not entirely sure this modification really changes the amount of data transferred to the GPU – or did you analyze this with Scout? (I honestly can't remember if Scout even shows that information.)

In any case: your second modification only works when having multiple bitmaps, and the speed difference is night and day according to your tests. So this is definitely the way to go! This is a fantastic enhancement, and I'm a little bit embarrassed that I never thought of trying that out. 🙉

I'll merge that right away! (Sorry for the delay - I'm currently really busy with other projects.)

PR #1121, opened 2025-03-18
Add vector drawing functions to Canvas

### LINKED ISSUE #1107: [Feature Suggestion] Vector Graphic Tesselation
Starling is the main way to use Hardware Acceleration on Flash, but currently, it does not natively support using Vector Art, who is Flash's main art form
My suggestion is adding tesselation support (converting it to triangles and feeding to the gpu), who is how all flash emulators (Scaleform, AwayFL, Ruffle) handle it, that way supporting using Vectors in hardware too.

Fancy2209:
Closes #1107 
Adds moveTo, lineTo and curveTo to starling.display.Canvas

Example
```as
package
{
	import flash.display.Shape;

	import starling.core.Starling;
	import starling.display.Sprite;
	import starling.display.Canvas;

	public class CanvasExample extends Sprite
	{

		public function CanvasExample()
		{
			super();
			var roundObject:Shape = new flash.display.Shape();
			Starling.current.nativeOverlay.addChild(roundObject);
			roundObject.graphics.beginFill(0x00FF00);
			roundObject.graphics.moveTo(250, 0);
			roundObject.graphics.curveTo(300, 0, 300, 50);
			roundObject.graphics.curveTo(300, 100, 250, 100);
			roundObject.graphics.curveTo(200, 100, 200, 50);
			roundObject.graphics.curveTo(200, 0, 250, 0);
			roundObject.graphics.endFill();
			roundObject.scaleX = roundObject.scaleY = 2;
			roundObject.x -= 250;

			var roundObjectStarling:Canvas = new Canvas();
			addChild(roundObjectStarling);
			roundObjectStarling.beginFill(0xFF0000);
			roundObjectStarling.moveTo(250, 0);
			roundObjectStarling.curveTo(300, 0, 300, 50);
			roundObjectStarling.curveTo(300, 100, 250, 100);
			roundObjectStarling.curveTo(200, 100, 200, 50);
			roundObjectStarling.curveTo(200, 0, 250, 0);
			roundObjectStarling.endFill();
			roundObjectStarling.scale = 2;
		}
	}
}
```
imagem

Not perfect, Starling's Mascot won't render (not properly at least, output with the current pr is nightmare fuel!) due to lack of Strokes/lineStyles (not versed enough to implement them, help is appreciated!) and the other more complex shape I tried (Away3D Logo) has some issues that make just some dots render
imagem
[CanvasVectorIssue.zip](https://github.com/user-attachments/files/19329777/CanvasVectorIssue.zip)

I still believe it's good enough for simpler shapes (as shown above), so decided to PR what I have.

### DIFF
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -171,7 +171,8 @@ package starling.display
         */
         public function drawGraphicsData(graphicsData:Vector.<IGraphicsData>):void
         {
-            for(var graphPropIndex:uint = 0; graphPropIndex < graphicsData.length; graphPropIndex++)
+            var graphicsDataLength:int = graphicsData.length;
+            for(var graphPropIndex:int = 0; graphPropIndex < graphicsDataLength; graphPropIndex++)
             {
                 var graphicsProperties:IGraphicsData = graphicsData[graphPropIndex];
 
@@ -183,7 +184,8 @@ package starling.display
                     var i:int = 0;
                     var data:Vector.<Number> = (graphicsProperties as GraphicsPath).data;
 
-                    for (var commandIndex:uint = 0; commandIndex < (graphicsProperties as GraphicsPath).commands.length; commandIndex++)
+                    var commandLength:int = (graphicsProperties as GraphicsPath).commands.length;
+                    for (var commandIndex:int = 0; commandIndex < commandLength; commandIndex++)
                     {
                         var command:int = (graphicsProperties as GraphicsPath).commands[commandIndex];
                         switch (command)
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -172,8 +172,9 @@ package starling.display
         public function drawGraphicsData(graphicsData:Vector.<IGraphicsData>):void
         {
             //var start:int = getTimer();
-            for each (var graphicsProperties:IGraphicsData in graphicsData)
+            for(var graphPropIndex:uint = 0; graphPropIndex < graphicsData.length; graphPropIndex++)
             {
+                var graphicsProperties:IGraphicsData = graphicsData[graphPropIndex];
                 if (graphicsProperties is GraphicsSolidFill)
                     this.beginFill(GraphicsSolidFill(graphicsProperties).color, GraphicsSolidFill(graphicsProperties).alpha);
                 else if (graphicsProperties is GraphicsPath)
@@ -181,8 +182,9 @@ package starling.display
                     var i:int = 0;
                     var data:Vector.<Number> = GraphicsPath(graphicsProperties).data;
 
-                    for each (var command:int in GraphicsPath(graphicsProperties).commands)
+                    for (var commandIndex:uint = 0; commandIndex < GraphicsPath(graphicsProperties).commands.length; commandIndex++)
                     {
+                        var command:int = GraphicsPath(graphicsProperties).commands[commandIndex]
                         switch (command)
                         {
                             case GraphicsPathCommand.MOVE_TO:
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -171,12 +171,13 @@ package starling.display
         */
         public function drawGraphicsData(graphicsData:Vector.<IGraphicsData>):void
         {
-            //var start:int = getTimer();
             for(var graphPropIndex:uint = 0; graphPropIndex < graphicsData.length; graphPropIndex++)
             {
                 var graphicsProperties:IGraphicsData = graphicsData[graphPropIndex];
+
                 if (graphicsProperties is GraphicsSolidFill)
                     this.beginFill(GraphicsSolidFill(graphicsProperties).color, GraphicsSolidFill(graphicsProperties).alpha);
+
                 else if (graphicsProperties is GraphicsPath)
                 {
                     var i:int = 0;
@@ -184,7 +185,7 @@ package starling.display
 
                     for (var commandIndex:uint = 0; commandIndex < GraphicsPath(graphicsProperties).commands.length; commandIndex++)
                     {
-                        var command:int = GraphicsPath(graphicsProperties).commands[commandIndex]
+                        var command:int = GraphicsPath(graphicsProperties).commands[commandIndex];
                         switch (command)
                         {
                             case GraphicsPathCommand.MOVE_TO:
@@ -208,13 +209,13 @@ package starling.display
                         }
                     }
                 }
+
                 else if (graphicsProperties is GraphicsEndFill)
                     this.endFill();
+
                 else
                     trace("[Starling] Canvas.drawGraphicsData: Unimplemented Graphics Data in input:", graphicsProperties, "at index", graphicsData.indexOf(graphicsProperties));
-
             }
-            // trace("Took " + (getTimer() - start) + " ms");
         }
 
         /** Removes all existing vertices. */
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -17,7 +17,7 @@ package starling.display
     import flash.display.GraphicsPathCommand;
     import flash.display.GraphicsEndFill;
     import flash.geom.Point;
-    
+
     import starling.geom.Polygon;
     import starling.rendering.IndexData;
     import starling.rendering.VertexData;
--- a/starling/src/starling/display/Canvas.as
+++ b/starling/src/starling/display/Canvas.as
@@ -15,6 +15,7 @@ package starling.display
     import starling.geom.Polygon;
     import starling.rendering.IndexData;
     import starling.rendering.VertexData;
+    import starling.utils.rad2deg;
 
     /** A display object supporting basic vector drawing functionality. In its current state,
      *  the main use of this class is to provide a range of forms that can be used as masks.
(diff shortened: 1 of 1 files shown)


Fancy2209:
Here's an example of how to read a flash shape with this
```as
for each(var command:int in graphicsProperties.commands)
                    {
                        switch(command)
                        {
                            case GraphicsPathCommand.MOVE_TO:
                                i = 0
                                this.moveTo(data[i], data[i+1])
                                i+=2
                                break
                            case GraphicsPathCommand.LINE_TO:
                                this.lineTo(data[i], data[i+1])
                                i+=2
                                break
                            case GraphicsPathCommand.CURVE_TO:
                                this.curveTo(data[i], data[i+1], data[i+2], data[i+3])
                                i+=4
                                break
                            default:
                                trace("Unimplemented Command", command)
                                break;
                        }
                    }
                } else if(graphicsProperties is GraphicsEndFill) 
                    this.endFill();
            }            
        }
```

Fancy2209:
The 2nd one could be due to how simple the Triangulation is, maybe I should look into Delaunay Triangulation for a follow up PR?

Fancy2209:
Passing the polygon trough the libtess2 CrossBridge port by Claus Whalers seems to have given a better result, so i really think this may be a triangulation issue
I'm not sure what the issue is though 
image

Fancy2209:
I was able to get the Away3D Logo rendering properly using libtess2.swc to tesselate the polygon and get it's Indices and Vertices:
image
```
protected function checkIfClosed():void
        {
            const curPath:Vector.<Number> = _paths[_paths.length - 1];
            
            const lastX:Number = curPath[curPath.length - 2];
            const lastY:Number = curPath[curPath.length - 1];
            
            if (lastX == curPath[0] && lastY == curPath[1])
            {
                _tess.newTess(1024*1024)
                _tess.addContour(curPath);
                _tess.tesselate(Tesselator.WINDING_ODD, Tesselator.ELEMENT_TYPE_POLYGONS);

                var polygon:Polygon = Polygon.fromVector(_tess.getVertices());
                var vertexData:VertexData = new VertexData();
                
                var numVertices:int = _tess.getVertexCount();
                var requiredTargetLength:int = numVertices;

                if (vertexData.numVertices < requiredTargetLength)
                    vertexData.numVertices = requiredTargetLength;

                for (var i:int=0; i<numVertices; i++)
                    vertexData.setPoint(i, "position", _tess.getVertices()[i*2], _tess.getVertices()[i*2+1]);

                vertexData.colorize("color", _fillColor, _fillAlpha);

                var indexData:IndexData = new IndexData(_tess.getElementCount());   
                for(var j:int = 0; j < _tess.getElementCount(); j++)
                    indexData.addTriangle(_tess.getElements()[j * 3], _tess.getElements()[j * 3 + 1], _tess.getElements()[j * 3 + 2])

                addChild(new Mesh(vertexData, indexData));
                _polygons[_polygons.length] = polygon;
                _tess.deleteTess();
            }
        }
 ```
 However, I don't think I can get an AS3 Port of libtess2 that could be vendored into starling...

**Comment by PrimaryFeather** [comment]
Oh wow, this is looking really cool – great work! 🤩

Even in the current state, it's already a great improvement on what the Canvas class can do – definitely worth adding!

Before I'm merging it, maybe you could look into a few small adaptations?

* Am I understanding it correctly that we always only need the last path in the `_paths` vector? If so, we don't necessarily need to store the previous ones, and just work with one. That would make it a little simpler and reduce memory consumption. (I know, polygons are kept - but that's because we need them for hit testing.)
* I think the method `checkIfClosed` could benefit from a better name. After all, it doesn't only check if the path if closed, but also draws the path it if it is. So maybe `drawPathIfClosed()` or something along those lines would be better?
* If you think it makes sense, we could also remove the `appendPolygon` method and move its logic into `drawPolygon`. I actually can't remember why I split that logic.

As for libtess2: yes, if there's a way to make use of the algorithm in question without having to add it as a dependency, that would be preferable – especially since Starling doesn't have any other dependencies at the moment, and this simplifies things for developers. But if it's just not feasible, we could still discuss this.

The license should not be an issue, in any case, except that we maybe need to include that SGI license with the source.

**Comment by PrimaryFeather** [comment]
A huge THANKS for making all those efforts, BTW! 🙌

**Comment by Fancy2209** [comment]
> A huge THANKS for making all those efforts, BTW! 🙌

No problem, was just kinda annoyed people got tesselation of flash vectors everywhere but in flash itself haha.

**Comment by Fancy2209** [comment]
By the way I don't have docs for the new functions as I wasn't sure what to add. I could copy them from the AS3 Docs but I wasn't sure where I should add the attribution (Adobe's Documentation is licensed under Creative Commons Non-Commercial with Attribution)

**Comment by nanjizal** [comment]
Fancy further to your pm, my GPU/pixel drawing experiments has been merged. 
My library will work in Starling but may need me to check after the refactor or you can pull a demo with any tweeks needed. But you can see current code here.
https://github.com/nanjizal/justGraphix/
currently there are no examples use cases, but general setup will be similar to these
https://github.com/nanjizal/cornerContourStarlingTest/blob/main/src/pi_xy/_application/TestStarling.hx ( pixels )
https://github.com/nanjizal/pi_xyStarlingTest/blob/main/src/pi_xy/_application/TestStarling.hx ( triangles )
In terms of tess requirements it handles thick/thin contours and even dash and a range of line ends, and of corners like mitre and fine but for a fill you can plug in several external libs, I can't remember if that is wired up ( but was in trilateral/3/cornerContour), as I was still considering if that part could be more flexible.
```
package justGraphix.contour3D.drawing;
// TODO: Not really happy with hard coding the triangulation possible engines.
enum abstract StyleFill( Int ) from Int to Int {
    var tess2;
    var polyK;
    var poly2tri;
}
```
these are haxelib's, polyK is basic so does not deal with special cases but fast. I ported that from js. tess2 is quite good but poly2tri is best for most complex shapes ( think ceramic uses variant ).
I do want to develop a triangulation lib as likely my contour stuff has much of the calculated info needed as it can do some cool gradient over width of line, so it can tell if a curve is going out or in.

It there is interest I can try to rig up starling demo and make sure all reworked code is compiling again, just I no longer work in code or with a computer so my down time is limited. Also I have not tested my lib against haxe flash target especially with pixelimage stuff. But lime peote nme openfl all have support

nanjizal:
the trouble with doing it all with triangles is that it can quickly use thousands, commerial libs like slug ( c++ ) and shapes ( unity ) do more in the shader rather then just rendering triangles

Fancy2209:
Ok so I made strokes render with a 0 alpha fill and this shape and was able to test the Starling animation
I could try to do a MovieClip but these speeds are slow

Libtess2
Took 26.348 seconds
imagem

Starling triangulation:
Took 9.088 seconds
imagem

PrimaryFeather:
@Fancy2209 Thanks a lot for implementing the additional changes, and for the performance test!

Wow, yeah, I wouldn't have expected it to take that much time, either. 😳
Still, it's definitely already a great addition, so it's definitely worth to be merged.

A short API-documentation for the new public methods would be great, you're right. Even if it's just very short. If I remember correctly, I often simply copied the docs from Adobe's API; since the Starling project was officially endorsed by Adobe (or no one ever checked 😆), it was never a problem. 

As for the unit tests ... well, I'm afraid most of the rendering code in Starling isn't unit tested, because I never took the time to create some kind of "golden image" logic (a big oversight on my side ...). So I guess it's okay if we skip those here, too.

Fancy2209:
Ok so asdoc added, also added drawGraphicsData as I've been using the function for testing already so might as well add it

Fancy2209:
I was using drawGraphicsData in the function I added but just read the Starling Guide and realized this isn't the best way to handle loops

Fancy2209:
Ok I think both PRs are ready now, though always open to changes

PrimaryFeather:
Thanks so much, @Fancy2209 – I just merged the changes into the master branch. 🎉

Fancy2209:
Nice! Any chance for #1123 next?

2jfw:
> @2jfw
> 
> > I've been using some rounded rectangles for buttons lately and was wondering if it would be a good addition to add a basic class for a RoundedRectangle making use of this, too?
> 
> Thanks for the suggestion, Jan! If I'm adding that, I'd rather think of a list of a whole bunch of useful shapes and add those altogether. In the meantime, I'm very glad that it should now be easy to create such a class for anyone with the new drawing methods. 😁

Yes, that's a good idea 👍

PrimaryFeather:
> Nice! Any chance for https://github.com/Gamua/Starling-Framework/pull/1123 next?

Done. 😁

Fancy2209:
I was reading the starling manual, doesn't this require section 2.10.2 to be updated?
https://manual.starling-framework.org/en/#_canvas_and_polygon

pol2095:
and the [doc](https://doc.starling-framework.org/current/starling/display/Canvas.html)

Fancy2209:
oh yeah the asdoc needs to be regenerated. I believe both things only daniel can do

PrimaryFeather:
Yes, that's on me! I typically update those when I make a new stable release. Which hasn't happened for a long time 😬 – but with the latest changes, it's high overdue. 👍

Fancy2209:
> Yes, that's on me! I typically update those when I make a new stable release. Which hasn't happened for a long time 😬 – but with the latest changes, it's high overdue. 👍 

So Starling 2.8 soon?

2jfw:
@PrimaryFeather: Can I ring a "possible" (?) short reminder bell on the issue from here and the follow up conclusions? https://github.com/airsdk/Adobe-Runtime-Support/issues/1773#issuecomment-1070993921 regarding Starling 2.8 ?

Fancy2209:
> Thanks for bringing the topic back to Harman's attention, @2jfw! 😁 
> 
> @Fancy2209 Yes, although what "soon" means is not yet sure. I'm really extremely busy in my job at the moment. 😬 But I'll try to get it done quickly. 

Is there anything that can be done to help with this?

PR #1123, opened 2025-03-21
Improve Polygon.triangulate() with Earcut

Fancy2209:
While working on #1121, I noticed Starling's polygon triangulation isn't very good (and sometimes surprisingly slow)
To combat this, I've ported https://github.com/mapbox/earcut/tree/v3.0.1, and made it the default, allowing the old algorithm to be used if so desired

Went from this:
imagem
Who took  ~8.9 seconds (8888 ms)

To this:
imagem
Who took 581 ms!


Fancy2209:
To achive the above I made  a modified version of MovieClip that uses removeChild and addChild to swap the frames (who are canvases) every frame, I'm willing to PR it, I just want to know, would it be best to add a new class or modify starling.display.MovieClip to work with Textures and Canvases?

Fancy2209:
Made tests also test earcut triangulation

PrimaryFeather:
Awesome, @Fancy2209! It's great that you added backward-compatibility with the optional property. Maybe we can remove the old algorithm altogether later, but it's good to be on the safe side.

Thanks also for adding unit tests as well! 🙏 

One quick question, though - could you maybe elaborate why you had to disable the earcut algorithm in one test? Did it produce a different result in that case? Thx!

Fancy2209:
> Awesome, @Fancy2209! It's great that you added backward-compatibility with the optional property. Maybe we can remove the old algorithm altogether later, but it's good to be on the safe side.
> 
> Thanks also for adding unit tests as well! 🙏
> 
> One quick question, though - could you maybe elaborate why you had to disable the earcut algorithm in one test? Did it produce a different result in that case? Thx!

The 2 triangulation tests have different outputs with and without earcut

PR #1144, opened 2026-02-25
MultiTextureStyle rework : fixed error when maxTextures > 2, up to 16…

MatseFR:
… textures + optimizations

There is currently an issue with `MultiTextureStyle` : when `MultiTextureStyle.maxTextures` is set to `3` or more the following error is thrown as soon as the program is activated :

`[Fault] exception, information=Error: Error #3647: AGAL validation failed: Temporary register read without being written to for source operand 1 at token 13 of fragment program.`

This version has been tested with 2000 movie clips with 1 to 16 different textures : I'm not sure I got the `programVariantName` right though ?

### DIFF
--- a/starling/src/starling/styles/MultiTextureStyle.as
+++ b/starling/src/starling/styles/MultiTextureStyle.as
@@ -456,7 +456,7 @@ class MultiTextureEffect extends MeshEffect
             }
             vertexFormat.setVertexBufferAt(3, vertexBuffer, "texture");
             context.setProgramConstantsFromVector(Context3DProgramType.FRAGMENT,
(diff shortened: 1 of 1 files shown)


Fancy2209:
If it's up to 16 textures shouldn't 
_MAX_NUM_TEXTURES be 16?

MatseFR:
@Fancy2209 it is, *if* the profile is not baseline, otherwise it's limited to 5 because of AGAL 1 (no ifs)

MatseFR:
all issues have been fixed, except the one that current MultiTextureStyle has too ( https://github.com/Gamua/Starling-Framework/issues/1145 )

cc @PrimaryFeather

PR #1148, opened 2026-07-16
Initialize _clippedViewPort to avoid null dereference on first render

mallyskies:
## Summary

`_clippedViewPort` is declared but never initialized in the `Starling` constructor, unlike its sibling `_previousViewPort`:

```as3
_viewPort = viewPort;
_previousViewPort = new Rectangle();
_stage = new Stage(viewPort.width, viewPort.height, stage.color);
```

It is only ever assigned inside `updateClippedViewPort()`, which `updateViewPort()` skips calling whenever `_viewPort` already equals the default-empty `_previousViewPort`. When that comparison happens to be true on the very first frame (depends on how/when the initial viewport is set relative to the first render call), `render()` goes on to dereference `_clippedViewPort.width` / `.height` while it is still `null`.

## How this was found

Discovered while porting a unit test suite to an openfl/starling (Haxe) project targeting js/html5. The Haxe port of Starling carries the identical gap, and it is reproducible there because html5's async Context3D creation, combined with a test harness that constructs/disposes many `Starling` instances back-to-back, hits the narrow first-frame window far more often than a normal, single-instance game bootstrap ever would. Tracing it back confirmed the same missing initialization exists here in the original AS3 source, so it predates the Haxe port entirely.

## Why this is an edge case

`updateViewPort()` only calls `updateClippedViewPort()` (the one place `_clippedViewPort` is ever assigned) when:

```as3
if (forceUpdate || !RectangleUtil.compare(_viewPort, _previousViewPort))
```

`initialize()` always calls `updateViewPort(true)` — forced, guaranteed to run `updateClippedViewPort()`. So under normal sequencing, `_clippedViewPort` is set once, immediately, before anything else can read it.

The bug only surfaces if `render()`'s own **unforced** call to `updateViewPort()` is the *first* one ever made for a given instance — i.e. something drives a render tick before `initialize()`'s forced call has run — **and** `_viewPort` happens to already equal the still-default `_previousViewPort` (both empty rects) at that moment. Once that coincidence happens, it doesn't self-correct: `_previousViewPort` is only ever updated inside that same skipped branch, so the guard keeps evaluating the same way on every subsequent frame, leaving `_clippedViewPort` null forever, not just for one frame.

Why it's rare in practice:
- It requires an ordering inversion between "GPU context ready" and the app's own forced init call. On native targets, context/window setup is effectively synchronous with app bootstrap, so this inversion basically can't occur.
- On html5, context creation is deferred (async), running alongside the browser's independent render-loop timing, which is what opens the window for this at all.
- A real game only ever passes through this narrow window **once**, at startup — even if the odds of hitting it on any given pass were non-trivial, one roll rarely comes up. The test harness that surfaced it constructs and disposes dozens of `Starling` instances back-to-back, re-rolling that same narrow window every time, which is what made a rare inversion actually reproducible.

Rather than staking the fix on fully nailing *why* the ordering can invert, the real defect is that `_clippedViewPort` has a representable-but-invalid state (`null`) that every other line of code assumes can't happen. The fix below removes that invalid state at construction, rather than trying to further constrain call ordering — a smaller, more robust fix regardless of the precise trigger.

## Fix

One line, mirroring the existing initialization of `_previousViewPort`:

```as3
_previousViewPort = new Rectangle();
_clippedViewPort = new Rectangle();
```

This only supplies a safe default; `updateClippedViewPort()` still recomputes it correctly whenever it actually runs.


PrimaryFeather:
Thanks a lot for the PR, @mallyskies! You're totally right, I must simply have overlooked this initialization. I merged that right away. 🙏

mallyskies:
Thanks Daniel!

